PR #1, opened 2026-05-13
Unify V1.1 and V1.2 firmware into a single file

Chris24680:
## Summary

- Replaces the two board-specific firmware directories (`Heltec V1.1/`, `Heltec V1.2/`) with a single unified source at `Pala_One_2_1/Pala_One_2_1.ino`
- Board selection is a one-line toggle at the top of the file — comment out `#define BOARD_V1_1` or `#define BOARD_V1_2` to match your hardware
- Uses a `using DisplayType = ...` alias so the rest of the ~4200-line file is untouched
- `partitions.csv` and `pala_one_sleep_black_icon_v4.h` were identical between both versions and are carried over unchanged
- `README.md` updated to point contributors to the new unified file

## Motivation

The two `.ino` files differed in exactly 3 lines (the display driver class name). Any bug fix or feature had to be applied twice, and it was easy to accidentally keep them out of sync.


Roanmh:
This is an important steps to making the code maintainable 👍 

✅ tested on v1.2 board

Pala_One_2_1/Pala_One_2_1.ino
```diff
@@ -1,5 +1,16 @@
...
+
+// ── Board selection: uncomment the line that matches your hardware ────────────
+#define BOARD_V1_1
+// #define BOARD_V1_2
+// ─────────────────────────────────────────────────────────────────────────────
+#ifdef BOARD_V1_1
+  using DisplayType = EInkDisplay_WirelessPaperV1_1;
+#else
+  using DisplayType = EInkDisplay_WirelessPaperV1_2;
+#endif
```
Roanmh, line 11: We could comment them both by default and give the user a helpful message if they forget. When I tested this PR, I immediately clicked "Deploy" with the wrong board type. It seemed to succeed, but the button didn't do anything on the device. Changing the board version fixed it. This would give the user feedback much more quickly, by failing the build, but with a nice message:

Tested:
- [x] Both commented results in nice error: `Compilation error: #error "Uncomment a board version"`
- [x] Either board un-commented compiles

```suggestion
// ── Board selection: uncomment the line that matches your hardware ────────────
#define BOARD_V1_1
// #define BOARD_V1_2
// ─────────────────────────────────────────────────────────────────────────────
#ifdef BOARD_V1_1
  using DisplayType = EInkDisplay_WirelessPaperV1_1;
  #define BOARD_CHOSEN 
#endif
#if defined BOARD_V1_2
  using DisplayType = EInkDisplay_WirelessPaperV1_1;
  #define BOARD_CHOSEN 
#endif
#ifndef BOARD_CHOSEN
  #error "Uncomment a board version"
#endif
```

Roanmh (changes requested):

PR #4, opened 2026-05-14
Native app loader: upload and run .bin apps via WiFi

Chris24680:
## Summary

- Adds a runtime app loader that loads, relocates, and executes position-independent `.bin` files stored in `/apps/` on LittleFS
- Apps are compiled separately and uploaded via the existing WiFi web interface — no firmware rebuild needed
- The Apps entry in the library menu discovers all `.bin` files in `/apps/` and lists them by name
- Adds `pala_app.h` and `pala_api.h` defining the binary header format and versioned function-pointer API (`PalaAPI`) that apps use to call back into the firmware
- Adds `examples/click_counter/` as a minimal reference app and `examples/palagotchi/` as a fuller example demonstrating persistence, timers, and multi-screen layouts
- Extends the WiFi upload page with an **Upload app (.bin)** card and a `/files` apps section with per-file delete buttons
- Removes the old `#ifdef APP_CLICK_COUNTER` in-firmware app stub; the click counter is now the uploadable example app
- Adds `examples/GETTING_STARTED.md` documenting the full app development workflow

## Technical notes

- Apps must be compiled `-fPIC -mlongcalls -fno-jump-tables -nostdlib` (Xtensa position-independent). `-fno-jump-tables` is mandatory: the loader only patches `.literal` pool relocations; jump tables emitted into `.rodata` are not patched and will crash the device
- The loader allocates executable SRAM via `heap_caps_malloc(MALLOC_CAP_EXEC | MALLOC_CAP_32BIT)`, maps the DIRAM DRAM view for data-bus access, applies `R_XTENSA_RELATIVE` relocations, then calls the entry point
- Apps are capped at 48 KB; the header carries a magic (`PALA`) and API version check so mismatched or corrupt files are rejected with an on-screen error
- `PalaAPI` is a frozen-order function-pointer struct — new fields are always appended and the version number bumped. Apps declare the version they were compiled against in their header; the loader rejects any mismatch
- **v1** — `clearScreen`, `drawHeader`, `drawTextAt`, `drawCenteredLarge`, `refreshDisplay`, `waitForEvent`, `snprintf_wrap`, `pollEvent`, `millisNow`, `buttonPressed`, `delayMs`, `pendingPresses`
- **v2** — adds `storageRead` / `storageWrite` (read/write `/apps/{key}.dat` on LittleFS) for persistent app state
- **v3** — adds `rtcSeconds()`, backed by `esp_rtc_get_time_us()`, which counts continuously through deep sleep. Used by palagotchi to apply stat decay correctly across sessions even after the device has slept
- `pendingPresses()` counts every individual short press-release unconditionally, bypassing the multi-click classifier — necessary because the e-ink refresh (~900 ms) is long enough for the triple-click window to accumulate four presses and trigger the quad-click discard

## Palagotchi

A Tamagotchi-style virtual pet included as a richer example app:

- Three stats — hunger, happiness, cleanliness — displayed as bar graphs; they decay over time (8h / 10h / 12h to empty respectively) whether the app is open or not
- Single click feeds, double click plays, triple click cleans; each action shows a brief animation screen before returning to the main view
- If any stat reaches zero the pet dies, showing a death screen with total alive time (wall-clock time from creation to death, including time the app was closed)
- Any button press on the death screen resets the pet and writes a fresh save
- Long press exits to the app launcher while the button is held, saving current state to LittleFS


ncont:
Good idea. 


One suggestion: add `on_single_pressed()`, `on_double_pressed()` and so on, API called and move the button handling to the base fw. I would suggest that the button recognition should be part of the main fw and not in the bin. 

Cheers, Noel

Chris24680:
Good point @ncont and thanks for the feedback, I did consider this approach. The challenge is that different app types suit different input models — an event-driven app like click counter would work well with callbacks, but palagotchi needs a game loop where it's do continuous work between button events (decay timers, display updates, action timeouts). Callbacks can technically support this too via an on_tick function, but it's an awkward inversion of control when the app naturally thinks in terms of "check timers, check input, update display, repeat".

It's worth noting that pollEvent does keep gesture recognition entirely in the main firmware — debouncing, multi-click timing, and gesture classification all live in the firmware's button handler. What apps receive is already a classified event code (single, double, triple, long), not raw GPIO state. The only thing the app controls is what it does in response to that event, which is necessarily app-specific behaviour.

This gives the same separation as callbacks — firmware owns recognition, app owns behaviour — but fits naturally whether the app is event-driven or loop-driven. The tradeoff is that click counter ends up with a slightly more verbose loop than it would with pure callbacks.

ncont:
I see. From briefly looking at your code, i thought that `app_main` is kind of of a `on_update`/`on_tick`method. Personally, i do like the callback approach as it creates cleaner function. 

And when thinking off multiple apps from different devs, the button handling can be implemented differently ( for example different long press ms), which would give the user a not unified behavior.

But anyway: nice implementation! Keep going.

PR #9, opened 2026-05-16
Add Samim 9pt font option and no-screensaver mode

quintenluyten2:
## Samim 9px font (size 9)
Adds `u8g2_font_samim_12_t_all` as a new 9px font size option in the font size selector. A few known limitations worth noting:

- Samim is a Persian/Arabic font. It will **not** render all Latin text correctly — glyphs will be missing or substituted with Arabic characters. It is intended for Persian/Arabic content.
- No bold variant exists in U8g2. `BOLD_FONT` is aliased to the same `samim_12_t_all` pointer as `MAIN_FONT` for this size.
- The encoding suffix is `_t_all` (transparent background, all Unicode) rather than `_te` (extended Latin) used by the other sizes.

## No-screensaver mode
Adds a checkbox toggle in Settings → Reading. When enabled:

- The device still goes to sleep on the normal timer.
- `drawSleepScreen()` is skipped — the last page stays visible on the e-ink display instead of being replaced by a screensaver image.
- On wake, the forced full-refresh cycle is skipped; the display resumes in fast-mode directly from the restored page. This allows the user to immediately start reading where they left off without waking for a full screen refresh

Persistent preference key: `cfg_noscr` (bool, default false).

### DIFF
--- a/Pala_One_2_1/Pala_One_2_1.ino
+++ b/Pala_One_2_1/Pala_One_2_1.ino
@@ -442,10 +442,18 @@ static void invalidateMetrics() {
   g_metricsValid = false;
 }
 
+// Returns the font ascent adjusted for layout purposes.
+// Samim 12 (size 9) reports an ascent 1 px taller than its visual cap
+// height warrants relative to the Helvetica faces. Subtracting 1 fits
+// an extra line on screen without the text feeling dense.
+static inline int layoutAscent() {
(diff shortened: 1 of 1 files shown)


quintenluyten2:
Not all changes in this PR have been thoroughly tested on the device yet. I will open review once I have tested it on device

PR #11, opened 2026-05-17
Hide Apps menu option when no apps are installed

tomascorreag:
## Summary

The Apps menu entry is currently rendered unconditionally in the library, even on a fresh device with no apps installed. Selecting it lands on an empty screen with no affordance to go back without context. This PR hides the Apps entry from `buildLibraryEntries()` when `/apps/` contains no `.bin` files.

## Behavior

- No `/apps/` directory, or `/apps/` exists but contains no `.bin`: Apps entry hidden.
- One or more `.bin` files in `/apps/`: Apps entry shown (unchanged from current behavior).
- Detection runs once per library rebuild, not on every render, so e-ink refresh cost is unchanged.

## Implementation

- New static `hasAnyApp()` helper: opens `/apps/`, returns true on the first file with a `.bin` extension. Closes the directory in all paths.
- Caller-site guard: `buildLibraryEntries()` only emits the Apps entry when `hasAnyApp()` is true.

`.ino` files get auto-generated forward declarations from the Arduino preprocessor, so the helper's definition below its call site is fine — both board variants compile clean.

### DIFF
--- a/Pala_One_2_1/Pala_One_2_1.ino
+++ b/Pala_One_2_1/Pala_One_2_1.ino
@@ -1333,7 +1333,7 @@ static void buildLibraryEntries() {
     g_library.entryDepths[g_library.entryCount] = 0;
     g_library.entryCount++;
   }
-  if (g_library.entryCount < MAX_LIBRARY_ENTRIES) {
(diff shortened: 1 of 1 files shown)


Chris24680:
I'm happy with this change, however would having a 'No apps installed' message on the apps page be better, so people know it's an option?

PR #12, opened 2026-05-17
Firmware hook to allow Sleep Timer and Streak Counter to work automatically

Higgs2209:
## Summary

Adds a Sleep Timer example app and firmware hooks that let it — and the
existing Reading Streak app — keep working when the user isn't in the
app (i.e. they're actually reading a book, or the device is in deep
sleep).

The design pattern: apps can't run in the background, so each app owns
a `/apps/<name>.dat` file and the firmware reads/writes that same file
when relevant reader events happen. Both sides duplicate the struct
layout, which becomes the wire-format contract.

### New: Sleep Timer app — `examples/sleep_timer/`
- Cycle preset durations via click (5/10/15/20/25/30/45/60 min).
- Double-click to start, hold to exit. Click during run cancels.
- State persists in `/apps/sleep_timer.dat` as 5 × `uint32_t`.

### Firmware integration — `Pala_One_2_1/Pala_One_2_1.ino`
A new ~130-line block defines:
- Wire-format structs (`ReadingStreakFile`, `SleepTimerFile`) — MUST
  match the example apps' layouts. Schema version field on both for
  future evolution.
- `readAppDat()` / `writeAppDat()` — thin LittleFS wrappers for
  `/apps/<key>.dat`.
- `streakAutoLogOnPageTurn()` — after `STREAK_PAGES_THRESHOLD` (5)
  page turns on a new day, writes `streak.dat` using the same log
  logic the app uses (bitmap shift, continuation check, etc.). Toast
  on success. Mirrors the app's algorithm exactly so manual taps and
  auto-log produce identical numbers.
- `sleepTimerCheckExpired()` — flips `RUNNING → NEEDS_NOTIFY` exactly
  once when the timer's end-time passes; toast at that moment. The
  app's own UI is the persistent indicator after that.
- `onReaderPageTurn()` — composite hook called after each real
  next/prev page turn.

Inserted call sites:
- 4 page-turn locations in the reader (next/prev × with/without
  save-throttle). Bookmark-add intentionally not hooked — re-renders
  but isn't a reading event.
- `setup()` calls `sleepTimerCheckExpired()` once so wake-from-RTC
  surfaces the toast immediately.
- `goToSleep()` arms `esp_sleep_enable_timer_wakeup(remainingUs)` if a
  timer is running, so the device wakes itself at the timer's end
  rather than waiting for the next button press.

### Reading Streak app — `examples/reading_streak/`
Daily check-in with rolling 32-day bitmap and streak/longest/total
stats. Manual click-to-log still works and writes the same file
format the firmware uses. Unchanged from prior reference work — the
new value is that you rarely need to open it.

### Tooling
- `examples/build.ps1` — Windows-native PowerShell builder. Uses the
  xtensa toolchain Arduino IDE installs at
  `%LOCALAPPDATA%\Arduino15\packages\esp32\tools\esp-x32\` plus your
  system Python. No `make` or Git Bash needed.
- `.gitignore` for `examples/**/*.bin` and `*.elf` artifacts.
- `examples/GETTING_STARTED.md` examples table updated.

## Tested on
- Heltec Wireless Paper **V1.2** (please test V1.1 too if you have one)
- Sleep timer in active reading: toast fires on next page turn after
  expiration.
-  Sleep timer deep-sleep wake: device wakes from RTC alarm, toast
  renders on the library/reader render that follows.
- Streak auto-log: streak advances after ≥5 page turns on a new day.
- Sleep timer cancel mid-run: returns cleanly to setup state.
-  Both apps compile to valid `PalaAppHeader` (verified via `build.ps1`).

## Tradeoffs / knobs
- `STREAK_PAGES_THRESHOLD = 5` — too low: idle browsing counts;
  too high: short sessions miss. Easy to tune.
- `TOAST_MS = 650` is the existing toast constant. If the sleep-timer
  toast feels too brief, happy to bump the global or add a
  longer-duration variant for just this case.
- Active RTC wake is a small power-profile change: the device boots
  once per expired timer, runs `setup()`, shows the toast, then
  returns to its normal idle-timeout deep-sleep. Negligible in practice.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,9 @@
+﻿################################################################################
+# This .gitignore file was automatically created by Microsoft(R) Visual Studio.
+################################################################################
+
+/.vs
+
+# Built app artifacts (per-example .bin/.elf produced by examples/build.ps1).
+examples/**/*.bin
+examples/**/*.elf
--- a/Pala_One_2_1/Pala_One_2_1.ino
+++ b/Pala_One_2_1/Pala_One_2_1.ino
@@ -4442,6 +4442,139 @@ static void drawSleepScreen() {
   display.update();
 }
 
+// ============================================================================
+//  App integration: reading-streak auto-log + sleep-timer notification
+//  Both features touch /apps/<key>.dat files that the corresponding example
+//  apps (examples/reading_streak, examples/sleep_timer) own. The structs
+//  below MUST match the SavedState/SleepTimerFile layouts in those apps —
+//  field order and types are the wire format.
+// ============================================================================
+
+struct ReadingStreakFile {
+  uint32_t version;        // == STREAK_SCHEMA
+  uint32_t firstRtcSec;    // rtcSeconds() at first ever write
+  uint32_t lastLoggedDay;  // day index, 0xFFFFFFFFu = never
+  uint32_t currentStreak;
+  uint32_t longestStreak;
+  uint32_t totalSessions;
+  uint32_t bitmapHead;     // day index of bit 0
+  uint32_t bitmap;         // bit i = "logged on day (head - i)"
+};
+
+struct SleepTimerFile {
+  uint32_t version;        // == SLEEP_TIMER_SCHEMA
+  uint32_t status;         // SLEEP_TIMER_IDLE / RUNNING / NEEDS_NOTIFY
+  uint32_t endRtcSec;
+  uint32_t durationMin;
+  uint32_t startedRtcSec;
+};
+
+static const uint32_t STREAK_SCHEMA          = 1;
+static const uint32_t STREAK_DAY_SECS        = 86400u;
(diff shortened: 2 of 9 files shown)


tomascorreag:
@Higgs2209 Heads up, the diff here is showing a lot more changes than I think you intended. It looks like this branch was created before #8 and #10 landed on main, so the diff is reverting them along with your sleep-timer / streak work, which does look correct.

A rebase onto current `main` should clean this up.

Higgs2209:
@tomascorreag Thank you! I believe I have cleaned everything up now. Hopefully it is suitable for main firmware.

Chris24680:
Does this overlap with the storageRead() and storageWrite() already implemented in the firmware?

Higgs2209:
Closinig as this was opened into /dev

PR #14, opened 2026-05-17
Modular refactor + PlatformIO support (Arduino IDE still works)

gnatpat:
I know this is a huge PR! It is essentially a full refactor (and a fair amount rewritten). I am not expecting this to be merged without discussion - I'm creating this PR to show and to start a [discussion in Discord](https://discord.com/channels/1502006584639881256/1505475742706831524). Very happy to discuss and change things.

Splits the monolithic Pala_One_2_1.ino into a layered module tree under Pala_One_2_1/src/ (pure/ -> storage/ -> hal/ -> ui/ -> web/).
Adds a PlatformIO build alongside the existing Arduino IDE flow. Same sources compile under both.

Does fix a number of bugs and changes a small amount of behaviour:
 - changing font size now correctly keeps position in book and bookmarks
 - uses light sleep in between main loop to save battery
 - long press now activates as soon as time has elapsed, rather than on release
 - toasts disappear automatically rather than on next draw
 - button presses rewritten to be a bit snappier, but still roughly the same time allowed
 - removed the LRU cache for page lookups in favour of just using the on disk one
 - adds maybe too many comments

Based on my rewrite here - https://github.com/gnatpat/e-reader

Usage patterns:
- Arduino IDE: open Pala_One_2_1/Pala_One_2_1.ino, uncomment BOARD_V1_1 or BOARD_V1_2 at the top, Verify/Upload. Same UX as before.
- PlatformIO: pio run -e wireless-paper-v1_2 -t upload (or -v1_1) - see instructions in README

Some changes were needed to allow the Arduino IDE to import and understand the new structure. I'm not sure how important we think keeping this existing behaviour is - especially with the web based flasher I can bring across in a next PR (uses GHA to build a firmware.bin and github hosted page to flash the reader - see https://github.com/gnatpat/e-reader and https://gnatpat.github.io/e-reader/).

pala_api.h and pala_app.h stay at the sketch root (their upstream location) so the existing examples/click_counter and examples/palagotchi Makefiles keep working with no edits.

Adds host-side CMake unit tests under test/ for the pure/codec/storage modules.

Verified locally:
  pio run -e wireless-paper-v1_2  -> SUCCESS (Flash 37.5%, RAM 36.8%)
  pio run -e wireless-paper-v1_1  -> SUCCESS (Flash 37.6%, RAM 36.8%)
  cmake --build test/build && ctest  -> passed
  Arduino IDE 2 (V1_2) Verify -> clean

Not yet tested on physical V1.1 as I don't have one. Would appreciate a tester.

## Suggested review flow

There is no diff as this explodes one file into loads. Some suggested reading:

1. `README.md` -> the *Include style* section explains why every `#include` has a `src/` prefix and why headers use `#ifndef` guards instead of `#pragma once`. Explains some of the set up weirdness caused by Arduino compatiability.
2. `Pala_One_2_1/Pala_One_2_1.ino` -> board-macro shim + `setup()`/`loop()` that replace the monolith. Should give you an idea of the screen based set up (current screen is passed a button, moving through screens is handled)
3. Take a look at one of the screens - the reader is the most important and most involved:
   - `src/ui/screens/reader_screen` -> defines top level screen, thin wrapper around deeper logic
   - `src/ui/reader` -> Functions/code around reading and mutating the currently open book
   - `src/hal/display` -> core UI/display code
   - `src/storage/book_metadata` -> storage of the book position and bookmarks
   - `src/pure/paginator`  -> pure pagination logic - has tests in test/test_paginator.cpp


Pala_One_2_1/Pala_One_2_1.ino
```diff
@@ -1,4500 +1,118 @@
...
+//        1. Install the Heltec ESP32 board package (heltec_wifi_lora_32_V3).
+//        2. Install libraries: heltec-eink-modules (todd-herbert fork),
+//           Adafruit GFX, U8g2_for_Adafruit_GFX.
+//        3. Uncomment exactly one of BOARD_V1_1 / BOARD_V1_2 below.
+//        4. Compile and upload.
+// ============================================================================
+
+// ── Board selection: uncomment the line that matches your hardware ──────────
+// #define BOARD_V1_1
+#define BOARD_V1_2
```
Roanmh: This should be commented in the repo to math the comment below and the README.


```suggestion
// #define BOARD_V1_2
```

Roanmh:
Want to share a transient error in case you see it.

After checking out your PR and building the project I got the error below, the error then resolved on subsequent runs. I wasn't able to repro with `pio run -t clean` and building again but I'm not familiar with any other caching/non-deterministic behavior of PIO, so idk if that's a good test:

```
*** [Pala_One_2_1\Pala_One_2_1.ino.cpp] TypeError : unsupported operand type(s) for -: 'float' and 'NoneType'
```

Roanmh:
I think there is an issue with the sleep changes:

1. upload to V1.2 board
2. allow device to sit on menu for 2 minutes
3. observe screensaver
4. observe almost immediate wake to menu again


https://github.com/user-attachments/assets/96df6b93-4a5c-4728-85b6-56aec7dfb5f0

gnatpat:
> I think there is an issue with the sleep changes:
> 
> 1. upload to V1.2 board
> 2. allow device to sit on menu for 2 minutes
> 3. observe screensaver
> 4. observe almost immediate wake to menu again
> 
>  PXL_20260517_170708912.TS.2.1.mp4

Thanks for the report! I had seen this a couple of times but thought it might be an issue with my board... I dug into this though and found a difference in the library versions used by Arduino and PlatformIO - after a rebuild sleep now works on my device!

> Want to share a transient error in case you see it.
> 
> After checking out your PR and building the project I got the error below, the error then resolved on subsequent runs. I wasn't able to repro with `pio run -t clean` and building again but I'm not familiar with any other caching/non-deterministic behavior of PIO, so idk if that's a good test:
> 
> ```
> *** [Pala_One_2_1\Pala_One_2_1.ino.cpp] TypeError : unsupported operand type(s) for -: 'float' and 'NoneType'
> ```

This one is interesting... I haven't seen this before - if it doesn't come back on clean I'd be tempted to ignore it for now unless we get more reports?

Roanmh:
>  after a rebuild sleep now works on my device!

✅ verified on my V1.2 device.

This makes me wonder about writing core dumps, or what ever the embedded equivalent could be.

> I'd be tempted to ignore it for now unless we get more reports?

✅ agreed. I think anything that could stop a beginner from getting it working should be taken seriously. But we would need a repro (or independent reports) to move forward.

README.md
```diff
@@ -21,11 +21,83 @@ There are currently two supported Heltec Wireless Paper versions:
...
+Firmware sources include each other by the full path from the sketch root, e.g. `#include "src/hal/display.h"`, not the shorter `#include "hal/display.h"`.
+
+This is to stay compatible with Arduino IDE. The IDE recursively compiles files inside the `src/` subfolder of a sketch, **but does not add `src/` to the compiler's include path** — only the sketch folder root is on it. So `#include "config.h"` from a file at sketch root fails (the file actually lives at `src/config.h`), whereas `#include "src/config.h"` resolves correctly. PlatformIO is happy either way; we picked the Arduino-IDE-compatible form so the same `#include` lines work in both build systems with no extra `-I` flags.
+
+## Host-side tests
+
+Pure modules and KV-backed storage have host-side unit tests under `test/`. They build with CMake and run on your laptop — no board required:
+
+```
+cmake -S test -B test/build && cmake --build test/build && ctest --test-dir test/build --output-on-failure
```
Roanmh: These instructions seem to be written in bash?

Maybe we should state that assumption. and mention that cmake and a compiler (gcc/g++?) are required. since this is separate from the platformio compiler, it seems worth mentioning.

Pala_One_2_1/src/web/files.cpp
```diff
@@ -0,0 +1,344 @@
...
+//  /            — landing page (storage card + upload form)
+//  /files       — file/folder browser
+//  /del         — POST: delete book
+//  /mkdir       — POST: create folder
+//  /rmdir       — POST: delete (empty) folder
+//  /move        — POST: move book to a different folder
+//  /jumppage    — POST: set the page that opens next on device
+// ============================================================================
+
+static void handleRoot() {
```
Roanmh, line 25: Should we consider a root file?

README.md
gnatpat: Good point - changed this to a redirect to the tests README which explains how to install and run!

Pala_One_2_1/src/web/files.cpp
gnatpat: I think it could make sense - would maybe suggest that could be a follow up rather than in this change?

Pala_One_2_1/src/ui/screens/upload_screen.h
```diff
@@ -0,0 +1,68 @@
...
+#include "src/state.h"   // File, FS
+#include "src/ui/screen.h"
+
+// ============================================================================
+//  Per-session state for a book / sleep-image upload. The struct is public
+//  so the upload-stream handlers in `web/web.cpp` can read/write it through
+//  `UploadScreen::state()`; only one upload session exists at a time, so
+//  storing it as a static member of UploadScreen is appropriate.
+// ============================================================================
+struct UploadState {
```
ncont: Personally, i would seperate the screen rendering from the Wifi state handling. Right now, the upload Screen does Wifi connection handling and drawing, which conflicts with the SRP.

This would be additionally helpfull as i plan to use BLE to connect to the device and then it would become messy to handle Wifi and BLE in the upload Screen.

gnatpat: Agreed - done!

PR #15, opened 2026-05-17
Reader menu and button mapping via webUI

Higgs2209:
## Summary

Adds a reader-menu overlay, a configurable status bar, three remappable
hold gestures, and a device-lock action — all controllable from the
existing web settings page. Single-file diff (firmware only), +279/-8.

### Reader-menu overlay (transient, in-reader)

A click-then-hold from the reader opens an info overlay that shows the
book title, current page, and percent through the book. Any other click
closes it and returns to the page. While open, a short click cycles the
statusbar mode in place (described below) so the user can experiment
without leaving the menu.

### Configurable statusbar (`cfg_statusbar`)

The bottom reader statusbar now has three modes:

- **Full** (default, 8px) — unchanged page-progress bar + page number.
- **Minimal** (1px) — single-row horizontal progress fraction at the
  bottom edge; saves 7px of vertical text area.
- **Hidden** (0px) — no statusbar reserve, full screen for text.

`STATUS_H` stays the constant 8 used by toast overlays so toast keeps
working regardless of mode. `getMetrics()` uses a new
`statusbarReserveH()` helper so the text-area line count tracks the
chosen mode after `invalidateMetrics()`.

### Three remappable hold gestures

Short / double / triple click bindings are unchanged (next page / prev
page / home). The three "hold" gestures are now user-bindable:

| Gesture                | Default   | Bindable to                      |
|------------------------|-----------|----------------------------------|
| Long press (≥850ms)    | Bookmark  | None / Bookmark / Lock / Menu    |
| Extra-long press (≥2s) | Lock      | None / Bookmark / Lock / Menu    |
| Click, then hold       | Menu      | None / Bookmark / Lock / Menu    |

Stored in `prefs` as `cfg_btnL` / `cfg_btnXL` / `cfg_btnCH`. Dispatcher
lives in `performReaderAction()`; the three hold paths in
`handleModeReader()` call it instead of hard-coding actions.

### Click-then-hold chord without losing single-click

A single short click already advances the page. To make
"click → press-and-hold" land as a single chord (rather than
"next page" followed by a long hold), `ButtonState::poll()` now defers
the pending `shortClick` emit while the button is still physically
pressed. On release:

- short release, `clickCount == 2` → `doubleClick` (prev page, as before)
- long release, `clickCount == 0` → `longClick` (e.g. bookmark)
- long release, `clickCount == 1` → `clickHoldClick` (the chord)

Single-click, double-click, triple-click, and plain hold all behave
identically to before; the suppression only kicks in while the button
is held.

### Device lock

Any gesture bound to **Lock** toggles `g_deviceLocked`. While locked,
the top of `loop()` drops every button event except the lock-bound
gesture, which toggles back to unlocked. A short toast confirms each
transition.

If no gesture is bound to Lock at all, the locked state is unreachable
— so the binding-set "all three holds set to non-Lock" can't strand the
user.

### Web UI

`/settings` gets a new **Buttons** card with three dropdowns, posted as
an independent form (the existing Reading card is untouched). The POST
handler clamps each value into the `ButtonAction` enum range and
persists. Defaults are written on first load, so reconfigured devices
behave exactly as before this PR.

## Tested on

- Heltec Wireless Paper V1.2
- Click-then-hold from reader → menu opens; statusbar cycle works.
- Extra-long press → device locks; same gesture unlocks; all other
  clicks are correctly swallowed while locked.
- Plain long press → bookmark, as before.
- Web `/settings` "Buttons" card → remap each gesture, reload page,
  selections persist, new bindings take effect immediately.
- Statusbar modes: Full / Minimal / Hidden all render correctly and
  re-flow the text area on change.

## Tradeoffs

- **`LONG_MS = 850`, `VERY_LONG_MS = 2000`** — same press-duration
  thresholds as today. Both are easy constants to tune.
- **Lock without a visible indicator** — the device shows whatever the
  reader was showing when locked. A small "🔒" badge could be drawn in
  a follow-up if it's worth the screen real estate.
- **Triple-click while locked** is also swallowed, so the "global
  triple-click → home" escape hatch is intentionally disabled in
  locked state. Lock gesture is the only way out.
- **`statusbarMode = HIDDEN` and toast** — toast still draws over the
  bottom 8px, briefly covering the lowest line of text. This matches
  current toast-vs-reader behaviour and seemed cheap to live with.

### DIFF
--- a/Pala_One_2_1/Pala_One_2_1.ino
+++ b/Pala_One_2_1/Pala_One_2_1.ino
@@ -2638,16 +2638,10 @@ static void drawReaderMenu() {
   u8g2.print(buf);
   y += lineH;
 
-  // Progress bar — full width, 8px tall.
-  int barX = MARGIN_X;
-  int barW = SCREEN_W - 2 * MARGIN_X;
-  int barH = 8;
-  int filled = (int)((pos * (uint32_t)barW) / (uint32_t)total);
-  if (filled < 0) filled = 0;
-  if (filled > barW) filled = barW;
-  gfx.drawRect(barX, y, barW, barH, 1);
-  if (filled > 2) gfx.fillRect(barX + 1, y + 1, filled - 2, barH - 2, 1);
-  y += barH + 4;
+  // Blank gap so the settings section reads as separate from progress info,
+  // not as a label for any visual above it. (The earlier full-width bar was
+  // visually indistinguishable from the bottom-of-screen statusbar.)
+  y += lineH;
 
   // Statusbar mode row — short-click cycles through the three modes.
   const char* sbName = "Full";
--- a/Pala_One_2_1/Pala_One_2_1.ino
+++ b/Pala_One_2_1/Pala_One_2_1.ino
@@ -4730,7 +4730,14 @@ void setup() {
             mode = MODE_READER;
             g_reader.pageTurnsSinceFull = FULL_REFRESH_EVERY_N_PAGES;
             renderCurrentPage();   // draw first — takes ~300ms, user releases button during this
-            resetInputFrontend();  // then discard the wake-press only
+            if (g_deviceLocked) {
+              // Keep the wake-press edges so a click-then-hold can wake AND
+              // unlock in one motion. resetInputFrontend would otherwise
+              // drain them and force the user to repeat the unlock gesture.
+              markUserActivity();
+            } else {
+              resetInputFrontend();  // discard the wake-press only
+            }
             restored = true;
           }
           break;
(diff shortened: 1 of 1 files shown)


ben-igel:
does lock only lock or send it to sleep?

Higgs2209:
> does lock only lock or send it to sleep?

@ben-igel No it doesn't, good pickup I over looked this. Let me fix it

Higgs2209:
Thanks @ben-igel  for pointing this out. It has now been fixed. Locking with now put the device to sleep and a click and long hold will wake it up and unlock it. You have to hold past the point the the device fully refreshes -I can't see a way around this.

Higgs2209:
Closing as now opened agaisnt /dev

PR #17, opened 2026-05-18
Stats app and relevant firmware changes

Higgs2209:
## Summary

Adds lifetime page-turn and button-press counters maintained by the
firmware, plus an example **Stats** app that displays them. Counters
live in RTC slow memory so deep-sleep cycles cost zero flash writes;
batched flushes to `/apps/stats.dat` keep the flash budget tiny
(~3500 writes/year for a heavy reader, vs ~10k–100k per-sector
lifetime).

### Firmware — stats subsystem (Pala_One_2_1.ino)

- New wire-format struct `StatsFile { version, firstRtcSec, pagesRead,
  buttonPresses }`. Must stay byte-identical to the matching struct
  in `examples/stats/app.c`.
- Working counters as `RTC_DATA_ATTR`:
  `g_statsPagesRead`, `g_statsButtonPresses`, `g_statsFirstRtcSec`,
  `g_statsEventsSinceFlush`, `g_statsRtcInit`. Survive deep sleep,
  reset on power-off.
- `statsEnsureLoaded()` — called once in `setup()` after `prefs.begin`.
  On cold boot it reads `/apps/stats.dat` and seeds the RTC counters;
  on deep-sleep wake (`g_statsRtcInit` still true from before sleep)
  it's a no-op.
- `statsBumpPages()` / `statsBumpButtons(delta)` — increment the RTC
  counter and bump an "events since last flush" tally. When that hits
  `STATS_FLUSH_EVERY_EVENTS` (100) a flush fires.
- `statsFlushToFile()` — writes the current totals to
  `/apps/stats.dat` via the existing `api_storageWrite` helper, then
  resets the events tally. Also called explicitly from `goToSleep()`
  so any in-flight deltas are persisted before the chip sleeps.
- `onReaderPageTurn()` hook — called from the four real reader
  page-turn sites (next/prev × with/without save-throttle in
  `handleModeReader` and `handleModeBookmarkPreview`).
  Bookmark-add is intentionally not hooked — it re-renders but isn't
  a real page advance.
- Button counting lives in `loop()`: `btns.rawPressCount` already
  increments unconditionally for every short press-release, so we
  just diff it against a `static uint32_t lastSeenPressCount` each
  tick and bump by the delta. Long-hold releases are not counted,
  which matches what users expect to see as a "click".

### App — `examples/stats/`

- `app.c` reads `/apps/stats.dat` once at entry via `api->storageRead`
  and displays:
  - "Pages read" (bold) + value
  - "Button presses" (bold) + value
  - "hold to exit" footer hint
- Long-press (≥850 ms) returns to the launcher, same convention as
  `click_counter`.
- Values are a snapshot at entry. While the app is running, the
  firmware `loop()` is paused, so any presses made while the screen
  is up don't update the displayed numbers until the app is reopened.
  Acceptable for a lifetime-counter readout; flagged here so reviewers
  don't expect a live ticker.

## Flash-wear math

`RTC_DATA_ATTR` storage is in the ESP32 RTC slow memory: unlimited
writes, survives deep sleep, lost on power-off. That's where the
working counters live.

Flush to `/apps/stats.dat` only happens:
- every 100 events (page turns + button presses combined), and
- at `goToSleep()`.

For a heavy reader doing 500 page turns + ~500 clicks per day
(1000 events) → 10 flushes/day → ~3650 writes/year on a single
LittleFS sector. ESP32 NOR flash budget is 10k–100k writes per
sector; this sits well inside the budget for the device's lifetime.

If the user yanks the battery, up to the last 100 events plus
anything since the most recent `goToSleep()` are lost. That's a
deliberate tradeoff to keep the flash quiet.

## Tested on

- Heltec Wireless Paper V1.2
- Page-turn from MODE_READER: lifetime counter increments correctly
  across both with-save and without-save sites; bookmark-add does
  not count.
- Bookmark-preview mode: page turns in the preview do count
  (matches the user-perceived behavior of "I turned a page").
- Button counting: every short press-release increments the counter
  by exactly 1; multi-click groupings (double/triple/quad) increment
  by N once, not separately.
- Deep sleep cycle: counters survive `esp_deep_sleep_start()` and
  resume from RTC RAM without a flash read.
- Cold boot (no `stats.dat` yet): app shows the default zeros, the
  file is created on first flush.
- Existing `stats.dat`: app reads it correctly and the firmware
  picks up where it left off.

## Tradeoffs 
- **No per-day / per-week breakdown** in this PR — `firstRtcSec` is
  stored so a follow-up could surface "days using" or "average
  pages/day" without a schema bump (`version` already reserved).
- **Stats are global, not per-book** — kept the data small and the
  hook surface tiny. A per-book variant would need a much larger
  storage scheme.

## Relationship to other open work

- **PR #12** (sleep_timer + reading_streak) also introduces an
  `onReaderPageTurn()` hook with the same four call sites. The two
  branches don't conflict at the base-commit level (both branch off
  `upstream/main`), but merging both produces a duplicate-function
  compile error. Whichever lands first, the other rebases and the
  stats subsystem either keeps just its bump-logic or drops the
  duplicate hook. Trivial conflict — ~5 lines of `onReaderPageTurn`
  body.

### DIFF
--- a/Pala_One_2_1/Pala_One_2_1.ino
+++ b/Pala_One_2_1/Pala_One_2_1.ino
@@ -4442,6 +4442,80 @@ static void drawSleepScreen() {
   display.update();
 }
 
+// ============================================================================
+//  Stats: lifetime page-turn + button-press counters
+//  Working counters live in RTC RAM (survive deep sleep, unlimited writes).
+//  Flash flush happens at sleep, library-root transition, and every
+//  STATS_FLUSH_EVERY_EVENTS bumps as a safety fallback. The wire-format
+//  struct below MUST match examples/stats/app.c.
+// ============================================================================
+
+struct StatsFile {
+  uint32_t version;        // == STATS_SCHEMA
+  uint32_t firstRtcSec;    // rtcSeconds() at first ever write
+  uint64_t pagesRead;
+  uint64_t buttonPresses;
+};
+
+static const uint32_t STATS_SCHEMA              = 1;
+static const uint32_t STATS_FLUSH_EVERY_EVENTS  = 100;
+
+RTC_DATA_ATTR uint64_t g_statsPagesRead       = 0;
+RTC_DATA_ATTR uint64_t g_statsButtonPresses   = 0;
+RTC_DATA_ATTR uint32_t g_statsFirstRtcSec     = 0;
+RTC_DATA_ATTR uint32_t g_statsEventsSinceFlush = 0;
+RTC_DATA_ATTR bool     g_statsRtcInit         = false;
+
+static void statsFlushToFile() {
+  StatsFile s;
+  s.version        = STATS_SCHEMA;
+  s.firstRtcSec    = g_statsFirstRtcSec;
+  s.pagesRead      = g_statsPagesRead;
+  s.buttonPresses  = g_statsButtonPresses;
+  api_storageWrite("stats", &s, (int)sizeof(s));
+  g_statsEventsSinceFlush = 0;
+}
+
+// Load from flash on cold boot (RTC RAM lost on power-off); on deep-sleep
+// wake the RTC counters are already authoritative.
+static void statsEnsureLoaded() {
+  if (g_statsRtcInit) return;
+  StatsFile s;
+  int n = api_storageRead("stats", &s, (int)sizeof(s));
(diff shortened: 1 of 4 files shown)


Higgs2209:
Closing as now opened agaisnt /dev

PR #18, opened 2026-05-18
Expose LoRa hardware to loadable apps via PalaAPI v4

Chris24680:
### Summary

  - Adds six raw LoRa primitives to PalaAPI (`loraInit`, `loraSleep`, `loraReady`, `loraRecv`, `loraSend`, `loraNodeId`), allowing loadable apps to drive the SX1262 radio directly
  - Introduces `lora_driver.h` — a minimal firmware-side hardware shim containing only what cannot live in an app: the RadioLib C++ objects, the IRAM ISR, and the receive queue
  - Adds `examples/mesh_chat/`, a loadable app that implements the full Meshtastic LongFast stack (AES-128-CTR, protobuf, packet assembly) entirely in position-independent C, with an inbox view and pre-set quick-message send menu

  ### Design

  The firmware owns only the irreducible hardware layer. All protocol logic — Meshtastic packet format, encryption, channel configuration — lives in the app. A different app could use the same LoRa primitives with a completely different protocol and parameters.

  ### Test plan

  - [ ] Firmware builds in Arduino IDE with RadioLib installed
  - [ ] `cd examples/mesh_chat && make` produces `mesh_chat.bin`
  - [ ] App appears in the Apps menu after uploading via WiFi
  - [ ] Serial monitor shows `[LORA] Ready — ...` on app launch
  - [ ] Long-pressing a message in the send menu logs `[LORA] TX N bytes`
  - [ ] Long-pressing in the inbox exits the app cleanly
  - [ ] Inbox persists across app restarts

  ### Help wanted

  I don't have a second Meshtastic device to test actual sending and receiving of messages over the air. If you have a Meshtastic node or a device running the Meshtastic app nearby, I'd really appreciate you testing it and reporting back — particularly whether received messages decrypt and display correctly, and whether sent messages appear on other nodes.


Pala_One_2_1/lora_driver.cpp
```diff
@@ -0,0 +1,103 @@
...
+    s_rxFlag = false;
+    size_t rxLen = s_radio.getPacketLength();
+    if (rxLen >= 4 && rxLen <= LORA_PKT_MAX) {
+        int nextHead = (s_rxHead + 1) % LORA_RX_QUEUE;
+        if (nextHead != s_rxTail) {
+            int state = s_radio.readData(s_rxQueue[s_rxHead].buf, rxLen);
+            if (state == RADIOLIB_ERR_NONE) {
+                s_rxQueue[s_rxHead].len = (int)rxLen;
+                s_rxHead = nextHead;
+            }
```
gnatpat, line 86: Does this silently drop packets if the queue gets full?

Chris24680: Yeah if the queue is full it skips the read data call. However the queue holds 4 packets and the app drains it every 10ms, so in normal use it's unlikely to fill up. 

I could add `Serial.printf("[LORA] RX queue full, packet dropped\n")` in the else branch to make it visible if that would be better? Or if you have any other ideas?

gnatpat: I think it's fine if it's a known constraint/choice - but yes either a serial print or a comment that mentions it explicitly would be good I think!

Chris24680: Done!

gnatpat (approved):

ncont (approved):
I can not check the Lora Part. But it builds .

PR #19, opened 2026-05-20
Skip saveList NVS write when stored bytes already match (restore PR #10 fix F)

tomascorreag:
## Summary

Restores the compare-before-write guard in `saveList` that was lost when the modular refactor (#14) pulled list persistence out of the monolith into `src/storage/list_items.cpp`. The monolith fixed this regression in #10 (fix "F"); the refactor brought the older buggy code forward and the guard never came with it.

`src/storage/list_items.cpp:15-19` — `saveList` writes to NVS unconditionally. The list screen calls `saveListItems` on every change AND on screen exit, so even an untouched list takes a flash-write hit every time the user opens and closes the list view.

```diff
 void saveList(KeyValueStore& kv, const ListData& data) {
   uint8_t buf[LIST_ENCODED_MAX_SIZE];
   size_t n = encodeList(data, buf);
+
+  // Skip the NVS write if the stored bytes already match. The list screen
+  // calls saveListItems on every change AND on screen exit; without this
+  // guard, untouched lists still take a flash-write hit on exit.
+  uint8_t existing[LIST_ENCODED_MAX_SIZE];
+  size_t got = kv.getBytes(LIST_KEY, existing, sizeof(existing));
+  if (got == n && std::memcmp(existing, buf, n) == 0) return;
+
   kv.putBytes(LIST_KEY, buf, n);
 }
```

Applied at the `KeyValueStore` boundary so both the firmware and host-test builds benefit.

## Tests

`test/test_list_store.cpp` gets one new case that asserts the no-op behavior precisely:

```cpp
TEST_CASE("saveList skips the write when the stored bytes already match") {
  ...
  saveList(kv, data);
  size_t writesBefore = kv.putBytesCallCount();
  saveList(kv, data);                              // same content
  CHECK_EQ(kv.putBytesCallCount(), writesBefore);  // no-op
  data.items[0].done = 1;
  saveList(kv, data);                              // real change
  CHECK(kv.putBytesCallCount() > writesBefore);    // wrote
}
```

`MapKvStore` gains a `putBytesCallCount()` accessor (additive — no change to existing tests).

**Verified locally** when first authored against gnatpat's branch: `cmake --build test/build && .\build\host_tests.exe` — **89 tests, 271 checks, 89 passed, 0 failed**, including the new case. This PR is a clean cherry-pick of that commit onto `dev`; no conflicts.

## Scope / history

Single commit. No public API changes.

Originally authored as gnatpat/pala-one-firmware#4 (against `refactor/platformio-modular` before #14 merged here). That PR bundled this fix with an ISR queue-race fix that gnatpat independently committed in `60ef3e0` — already on this branch — so only the `saveList` half is re-submitted.

### DIFF
--- a/Pala_One_2_1/src/storage/list_items.cpp
+++ b/Pala_One_2_1/src/storage/list_items.cpp
@@ -15,6 +15,14 @@ ListData loadList(KeyValueStore& kv) {
 void saveList(KeyValueStore& kv, const ListData& data) {
   uint8_t buf[LIST_ENCODED_MAX_SIZE];
   size_t n = encodeList(data, buf);
+
+  // Skip the NVS write if the stored bytes already match. The list screen
+  // calls saveListItems on every change AND on screen exit; without this
+  // guard, untouched lists still take a flash-write hit on exit.
+  uint8_t existing[LIST_ENCODED_MAX_SIZE];
+  size_t got = kv.getBytes(LIST_KEY, existing, sizeof(existing));
+  if (got == n && std::memcmp(existing, buf, n) == 0) return;
+
   kv.putBytes(LIST_KEY, buf, n);
 }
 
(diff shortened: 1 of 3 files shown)


ncont:
Tested with v1.2. No issues found.

PR #20, opened 2026-05-20
Add build-time language selection (English / Spanish-LA)

tomascorreag:
## Summary

Adds build-time language selection to the firmware using a Tasmota/Marlin-style preprocessor mechanism. One language per binary; selected by a `LANG_*` build flag. All on-device and captive-portal UI strings now go through `D_*` macros (~246 keys), concatenated adjacent-literal at the call site — zero runtime cost, binary byte-identical to the prior inline literals under `LANG_EN`.

Two languages ship:
- `LANG_EN` — English (default; identical to pre-PR output)
- `LANG_ES_LA` — Spanish (Latin America)

## How it's selected

| Build system | How to pick the language |
|---|---|
| **PlatformIO** | New leaf envs combine board × language via `extends`: `wireless-paper-v1_{1,2}-{en,es}`. The bare board envs remain as the base. |
| **Arduino IDE** | Uncomment `LANG_EN` or `LANG_ES_LA` near the top of `Pala_One_2_1.ino`, alongside the `BOARD_V1_X` block. |

If neither macro is set, `src/lang/lang.h` defaults to `LANG_EN` with a `#pragma message` — so host CMake tests and uncustomised Arduino IDE builds still compile.

## Where strings live

```
Pala_One_2_1/src/lang/
  lang.h     # selector + LANG_EN fallback
  en.h       # canonical key set (~246 D_* defines, sectioned by surface)
  es_la.h    # Spanish-LA mirror — every key present
```

`en.h` is the source of truth for the key set. A missing key in `es_la.h` produces a compile error against the `-es` build (undeclared identifier), so drift is structurally detected. Authoring rules — placeholders, HTML entities, JS-confirm quoting — are documented at the top of `lang.h`.

## Glyph coverage

Spanish-LA needs `á é í ó ú ñ Ñ ¿ ¡ ü Ü`. Body / bold / app-large / toast roles use the existing `u8g2_font_helv*_te` Latin Extended tables. A new `Font::useToast()` role splits toasts off `useUiSmall` so translated toasts can render accents. `useUiSmall` (battery %) and `useUiTiny` (page number) intentionally stay on ASCII-only `_tf` tables — they only render digits/% and a switch would resize the battery indicator. Web responses already declare `charset=utf-8`.

## Out of scope

- Runtime language switching (build-time-only by design).
- Languages outside Latin Extended (Cyrillic / CJK would require different u8g2 fonts — separate decision).
- CI matrix to build all 4 binaries on tag — project has no CI today.

## Incidental

Added `#include <ostream>` to `src/pure/arduino_compat.h` — fixes a pre-existing MSVC host-test build error where the friend `operator<<(std::ostream&, const String&)` was declared without the complete type in scope. No-op for GCC/Clang (libstdc++ pulled it in transitively via `<string>`); host-only branch — firmware build is unchanged.

## Test plan

- [x] All four PIO envs build cleanly (`wireless-paper-v1_{1,2}-{en,es}`)
- [x] Host CMake tests pass (`ctest --test-dir test/build`) — `test_bookmarks_codec.cpp` updated to compare against `D_TOAST_BOOKMARK_EXISTS` (language-agnostic) with an `#ifdef LANG_EN`-gated literal pin
- [x] Flashed `wireless-paper-v1_2-es` to hardware and walked: library / reader / bookmarks (toast strings) / about / upload / apps / list / app-loader error overlays — no missing-glyph squares, no clipping
- [x] Walked all captive-portal routes on Spanish build via SoftAP (`PALA-XXXXXX` / `palaread` → `192.168.4.1`): `/`, `/files`, `/bookmarks`, `/list`, `/settings`, `/upload-app`, `/reset` — accents render correctly, JS confirm dialogs in Spanish, bookmark export `.txt` carries Spanish labels
- [ ] English build (`wireless-paper-v1_2-en`) walk-through regression spot-check — recommended before merge

## Notes for reviewers

- External app ABI (`pala_api.h`, `pala_app.h`, `PALA_API_VERSION = 3`) is unchanged.
- Adjacent-literal concat means the LANG_EN build produces output byte-identical to the prior inline-literal code at every call site — no behaviour change for the default language.
- Adding a third language is purely additive: clone one of the headers, add an `#elif defined(LANG_*)` arm to `lang.h`, add two leaf envs in `platformio.ini`.

### DIFF
--- a/Pala_One_2_1/Pala_One_2_1.ino
+++ b/Pala_One_2_1/Pala_One_2_1.ino
@@ -28,6 +28,14 @@
 // #define BOARD_V1_2
 // ────────────────────────────────────────────────────────────────────────────
 
+// ── Language selection: uncomment exactly one (Arduino IDE) ─────────────────
+//   PlatformIO users pick the env in platformio.ini (-en / -es leaf envs)
+//   and can leave these defines commented out. Default if nothing is set:
+//   English (with a #pragma message warning from src/config.h).
+// #define LANG_EN
+// #define LANG_ES_LA
+// ────────────────────────────────────────────────────────────────────────────
+
 // When built with PlatformIO, WIRELESS_PAPER + DISPLAY_V1_x come from
 // build_flags and the BOARD_V1_x macros above stay commented out. When
 // built with Arduino IDE, the macros above drive the same defines so the
@@ -124,7 +132,7 @@ void setup() {
   display.clear();
 
   if (!fsBegin()) {
-    drawCenter("Storage error", "Try factory reset");
+    drawCenter(D_BOOT_STORAGE_ERROR, D_BOOT_TRY_FACTORY_RESET);
     return;
   }
   ensureBooksDir();
--- a/Pala_One_2_1/Pala_One_2_1.ino
+++ b/Pala_One_2_1/Pala_One_2_1.ino
@@ -32,7 +32,7 @@
 //   PlatformIO users pick the env in platformio.ini (-en / -es leaf envs)
 //   and can leave these defines commented out. Default if nothing is set:
 //   English (with a #pragma message warning from src/config.h).
-// #define LANG_EN
+#define LANG_EN
 // #define LANG_ES_LA
 // ────────────────────────────────────────────────────────────────────────────
 
--- a/Pala_One_2_1/src/pure/arduino_compat.h
+++ b/Pala_One_2_1/src/pure/arduino_compat.h
@@ -18,7 +18,7 @@
 #include <cstdint>
 #include <cstdio>
 #include <cstring>
-#include <ostream>   // operator<<(std::ostream&, const String&) below
+#include <ostream>
 #include <string>
 
 // Minimal Arduino `String` shim. Only the methods used by the pure modules
--- a/Pala_One_2_1/src/pure/arduino_compat.h
+++ b/Pala_One_2_1/src/pure/arduino_compat.h
@@ -18,6 +18,7 @@
 #include <cstdint>
 #include <cstdio>
 #include <cstring>
+#include <ostream>   // operator<<(std::ostream&, const String&) below
 #include <string>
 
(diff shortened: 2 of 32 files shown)


Pala_One_2_1/src/pure/arduino_compat.h
```diff
@@ -18,6 +18,7 @@
 #include <cstdint>
 #include <cstdio>
 #include <cstring>
+#include <ostream>   // operator<<(std::ostream&, const String&) below
```
ncont: i think ostream is well known and we dont need a comment here.

tomascorreag: Agreed, just removed the comment.

Pala_One_2_1/Pala_One_2_1.ino
```diff
@@ -28,6 +28,14 @@
 // #define BOARD_V1_2
 // ────────────────────────────────────────────────────────────────────────────
 
+// ── Language selection: uncomment exactly one (Arduino IDE) ─────────────────
+//   PlatformIO users pick the env in platformio.ini (-en / -es leaf envs)
+//   and can leave these defines commented out. Default if nothing is set:
+//   English (with a #pragma message warning from src/config.h).
+// #define LANG_EN
```
ncont: I would leave the define uncommented. When useres want to start they should have a low barrier to start and not be scared by a warning.

ncont:
Tested with Debian and Board v1.2:

Build with PlatformIO:

- [x]  English
- [x] Spanish

Build with Arduino:

- [x]  English
- [x] Spanish

Walked both languages and could not find any severe issue ( i can not read spanish tho)

Pala_One_2_1/Pala_One_2_1.ino
tomascorreag: Sounds good, fixed.

PR #25, opened 2026-05-22
Dev/sleep streak hooks

Higgs2209:
## Summary

Re-target of #12 against the new `dev` (PlatformIO modular) layout. The original main-targeted patch won't apply after the refactor, so this is a fresh port; on-disk wire format is identical to the original.

Two example apps plus the firmware hooks they need to keep working when the user is in the book rather than the app:

- **`examples/reading_streak/`** — daily check-in with a rolling 32-day bitmap and streak / longest / total stats. Persists `/apps/streak.dat`.
- **`examples/sleep_timer/`** — cycles preset durations (5/10/15/20/25/30/45/60 min) via click; double-click starts, hold exits, click during run cancels. Persists `/apps/sleep_timer.dat`.
- **`examples/build.ps1`** — Windows PowerShell builder for the example apps (no `make` / Git Bash needed). Firmware still builds via PlatformIO; this is only for the app `.bin` builds.

The firmware side splits into two pieces. The deterministic core — bitmap-shift + continuation rules that decide what a "log session for day N" does to the saved state — lives in a pure module so it's host-testable:

**`Pala_One_2_1/src/pure/streak_log.{cpp,h}`**

```
cpp
StreakLogResult applyStreakLog(const ReadingStreakFile& current, uint32_t today);
```

The I/O wrapper (read/write /apps/streak.dat, bootstrap on first call, pages-today threshold gate) lives in Pala_One_2_1/src/ui/reading_hooks.cpp alongside the sleep-timer hooks:

- onReaderPageTurn() is called from advancePage() / retreatPage() in src/ui/reader.cpp. Threshold-gates and then delegates the state transition to applyStreakLog. Bookmark-add doesn't go through advance/retreat, so it correctly stays uncounted.

- sleepTimerCheckExpired() flips a running timer to NEEDS_NOTIFY exactly once when its end-time passes; called from setup() so a timer that expired during deep sleep surfaces its toast on the first render after wake.

- sleepTimerWakeUs() is consulted by Sleep::enter() to arm esp_sleep_enable_timer_wakeup() if a timer is running, so the device wakes itself at the timer's end-time rather than waiting for the next button press.

The struct layout in streak_log.h mirrors the app-side SavedState byte-for-byte, with a schema-version field gating future upgrades.

No .gitignore changes are needed — dev's blanket *.bin / *.elf rules already cover the example apps' build artifacts.

Tests
test/test_streak_log.cpp — 7 cases exercising the pure rules against concrete day indices, no FS / RTC / Toast needed:

- First-ever log starts streak at 1, longest = 1, totalSessions = 1.
- Consecutive days extend streak; longest tracks the peak.
- Gap (today > lastLoggedDay + 1) resets currentStreak to 1, preserves longestStreak.
- Same-day re-log is a no-op (changed=false).
- Bitmap shift within the 32-day window leaves the expected bits set (e.g. days 100 → 101 → 103 = 0b1101 from head 103).
- Bitmap shift ≥ 32 days clears the bitmap — applyStreakLog clamps the shift explicitly so a uint32_t << >=32 (UB in C++) can't manifest.
- firstRtcSec + version pass through unchanged.

Verified locally on the rebased branch (Windows / MSVC 19.50, all 7 cases + 113 pre-existing tests pass, 0 failures):
```

cmake -S test -B test/build
cmake --build test/build
```

### DIFF
--- a/Pala_One_2_1/Pala_One_2_1.ino
+++ b/Pala_One_2_1/Pala_One_2_1.ino
@@ -77,6 +77,7 @@
 #include "src/ui/font.h"
 #include "src/ui/pala_api_impl.h"
 #include "src/ui/reader.h"
+#include "src/ui/reading_hooks.h"
 #include "src/ui/screen.h"
 #include "src/ui/widgets.h"  // drawCenter
 #include "src/ui/screens/about_screen.h"
@@ -151,6 +152,10 @@ void setup() {
   initPalaAPI();
   registerWebRoutes();
   markUserActivity();
+
+  // If a sleep timer expired (possibly woke us via RTC alarm), queue the
+  // toast so it appears on whatever screen we render next.
+  sleepTimerCheckExpired();
 
   if (tryRestoreReadingSession()) {
     renderCurrentPage();      // ~300ms draw — wake-press releases during this
--- a/Pala_One_2_1/src/pure/streak_log.cpp
+++ b/Pala_One_2_1/src/pure/streak_log.cpp
@@ -1,0 +1,30 @@
+#include "src/pure/streak_log.h"
+
+StreakLogResult applyStreakLog(const ReadingStreakFile& current, uint32_t today) {
+  StreakLogResult r{};
+  if (current.lastLoggedDay == today) {
+    r.changed = false;
+    r.next    = current;
+    return r;
+  }
+
+  ReadingStreakFile s = current;
+
+  if (today > s.bitmapHead) {
+    uint32_t d = today - s.bitmapHead;
+    s.bitmap     = (d >= 32) ? 0u : (s.bitmap << d);
+    s.bitmapHead = today;
+  }
+  s.bitmap |= 1u;
+
+  bool cont = (s.lastLoggedDay != STREAK_DAY_UNSET)
+           && (today == s.lastLoggedDay + 1);
+  s.currentStreak  = cont ? (s.currentStreak + 1) : 1;
+  if (s.currentStreak > s.longestStreak) s.longestStreak = s.currentStreak;
+  s.lastLoggedDay  = today;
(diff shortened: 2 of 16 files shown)


gnatpat:
Unless I'm misunderstanding, the way the streak information is transferred between the firmware and the app is that we write a struct to a file, and then the app has the same struct and reads it?
That seems like a lot of implicit shared connections and internal knowledge an app now needs to have. It also doesn't match the existing pattern of apps only interacting through the api. I would be keen to keep that pattern and expose this though the API? That also comes with the benefit that if we need to change the struct the API version is bumped and apps explicitly stop working rather than fail with an error when reading.

I also don't really understand the sleep timer? Could you explain what it is and what it does?

gnatpat:
I guess I even wonder if the streak view should be an app at all or if it should just be part of the firmware?

ncont:
I would copy my comment from the pr #26 (https://github.com/PaulLagier/pala-one-firmware/pull/26#issuecomment-4520529217) to here and share @gnatpat concern. Since this is a constant running module in the FW, it should have the view in it as well and not separate as an app. 

Maybe it make sense to have a "Statistics" Screen and plot the streak and Lifetime info's together?

Higgs2209:
@gnatpat @ncont 

I was initially going to do both the streaks and stats as part of the firmware, but wanted to use the app systems as I assumed this was the better way to dev going forward.

I have taken on and agree with your feedback on this a #26 and will rework and open new PRs.

Thank you both - Still new so I appreciate the patience.

ncont:
I will put this and #26 in draft mode. You can simple modify them. No need to reopen again. As soon as you are ready, mark them as ready and we have a second look.

ncont:
Merged with #45

PR #26, opened 2026-05-22
Lifetime page-turn and button-press counters + Stats example app

Higgs2209:
## Summary

Re-target of #17 against the new `dev` (PlatformIO modular) layout. Same wire format and behaviour as the main-targeted PR.

A lifetime-counter subsystem so apps can read totals without the firmware hammering flash on every event, plus a Stats example app that displays them.

**`Pala_One_2_1/src/storage/stats.{cpp,h}`** owns the working counters in RTC RAM (survive deep sleep with zero flash writes) and the flash I/O. The wire-format struct + schema-version check live in a pure helper so they're host-testable:

**`Pala_One_2_1/src/pure/stats_codec.{cpp,h}`**

```cpp
struct StatsFile { uint32_t version; uint32_t firstRtcSec; uint64_t pagesRead; uint64_t buttonPresses; };
size_t encodeStats(const StatsFile& s, uint8_t* buf);
bool   decodeStats(const uint8_t* buf, size_t len, StatsFile& out);
```

Public API on the storage side:

- `statsBumpPages()` — called from `advancePage()` / `retreatPage()` in `src/ui/reader.cpp` so only real cursor moves count.
- `statsBumpButtons(delta)` — called from `loop()` against a diff of `g_btns.peekPressCount()`.
- `statsFlushToFile()` — called from `Sleep::enter()` so any in-flight deltas land before power-down. Also auto-flushes every `STATS_FLUSH_EVERY_EVENTS` (= 100) bumps as a safety net. A heavy 500-turn/day reader is ~5 flash writes/day.

`src/hal/input.h` gains a `peekPressCount() const` accessor alongside the existing `consumePressCount()`. Stats reads via peek (doesn't disturb the apps-API consumer); if the apps API consumed between ticks (current < lastSeen), stats clamps and continues without underflow.

**`examples/stats/`** reads `/apps/stats.dat` and shows total pages + total button presses. Long-press to exit. Values are snapshotted at app entry — the firmware `loop()` is paused while the app runs, so any clicks made while reading the screen don't update the on-screen numbers until you re-open the app.

`StatsFile` in `stats_codec.h` must stay byte-identical to the struct in `examples/stats/app.c`.

## Tests

`test/test_stats_codec.cpp` — **5 cases**:

- Round-trip through `encodeStats` + `decodeStats` preserves every field.
- Wrong `version` is rejected (a future schema bump must reinitialise from zero, not return stale fields).
- Short buffer is rejected.
- Oversized buffer is rejected — a future format growth must not silently slice off the first `STATS_ENCODED_SIZE` bytes and call it valid.
- `STATS_ENCODED_SIZE == sizeof(StatsFile)` — guards against accidental padding drift.

Verified locally on the rebased branch (Windows / MSVC 19.50, all 5 cases + 115 pre-existing tests pass, 0 failures):

```powershell
cmake -S test -B test/build
cmake --build test/build
```

## Scope / history

Two commits — the original feature commit plus the pure-codec extract + tests. Rebased on current `dev` (post-PR #20 language selection); auto-merge clean. No public API changes outside the additive `peekPressCount` and the new `stats.h` / `stats_codec.h` modules.

If the sleep-streak-hooks PR merges first, conflicts in `src/ui/reader.cpp`, `src/ui/sleep.cpp`, `Pala_One_2_1/Pala_One_2_1.ino`, and `test/CMakeLists.txt`. Resolution is "keep both calls" — the features are independent.

### DIFF
--- a/examples/stats/app.c
+++ b/examples/stats/app.c
@@ -1,0 +1,79 @@
+#include "../../Pala_One_2_1/pala_app.h"
+#include "../../Pala_One_2_1/pala_api.h"
+
+__attribute__((section(".header")))
+const PalaAppHeader pala_header = {
+    .magic        = PALA_APP_MAGIC,
+    .api_version  = PALA_API_VERSION,
+    .name         = "Stats",
+    .entry_offset = 0,  // patched by Makefile
+    .reloc_offset = 0,  // patched by Makefile
+    .reloc_count  = 0,  // patched by Makefile
+};
+
+#define LONG_PRESS_MS 850
+#define STATS_SCHEMA  1u
+
+// Must match StatsFile in Pala_One_2_1.ino byte-for-byte.
+typedef struct {
+    uint32_t version;
+    uint32_t firstRtcSec;
+    uint64_t pagesRead;
+    uint64_t buttonPresses;
+} StatsFile;
+
+// snprintf in the app runtime may not advertise %llu support; render
+// values >= 2^32 as a two-segment decimal to avoid relying on it.
+static void formatU64(const PalaAPI* api, char* buf, int len, uint64_t v) {
+    if (v <= 0xFFFFFFFFu) {
+        api->snprintf_wrap(buf, len, "%lu", (unsigned long)v);
+        return;
+    }
+    unsigned long hi = (unsigned long)(v / 1000000000ULL);
+    unsigned long lo = (unsigned long)(v % 1000000000ULL);
+    api->snprintf_wrap(buf, len, "%lu%09lu", hi, lo);
+}
+
+static void drawPage(const PalaAPI* api) {
+    StatsFile s;
+    int n = api->storageRead("stats", &s, (int)sizeof(s));
+    int valid = (n == (int)sizeof(s) && s.version == STATS_SCHEMA);
+
+    api->clearScreen();
+    api->drawHeader("Stats");
+
+    char num[32];
+
+    api->drawTextAt(6, 36, "Pages read", 1);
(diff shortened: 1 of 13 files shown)


ncont:
Hey, 

i like the idea! Altho, i would say the display of the stats can go directly in the Firmware and not in an App. It is so dependent on the FW. And to tell a user that they need an app to load onto the device just to see the stats even though they are collected all the time would be a bit complicated. At least for me :D

That would also eliminate the possibility that a user loads a old app into a new FW where the stats struct changed. 

So, i would vote to have a "Statics" entry in the main menu. 

What do you think?

examples/stats/app.c
```diff
@@ -0,0 +1,79 @@
...
+    int n = api->storageRead("stats", &s, (int)sizeof(s));
+    int valid = (n == (int)sizeof(s) && s.version == STATS_SCHEMA);
+
+    api->clearScreen();
+    api->drawHeader("Stats");
+
+    char num[32];
+
+    api->drawTextAt(6, 36, "Pages read", 1);
+    if (valid) formatU64(api, num, sizeof(num), s.pagesRead);
```
ncont, line 48: I think a direct info about the validity is better that setting them to "--". 

Like "Invalid statistic format detected". And in that case: is there any way to clean them up?

Higgs2209:
@ncont  - I have updated this to a new branch "statistics-screen"  - shows button presses, streaks as a bit map because I didn't see this comment....

Sorry - Should I change this structure to work with this existing PR?

ncont:
In that case open a new one and close the old ones .

ncont:
Merged with #45

PR #27, opened 2026-05-22
 Reader menu overlay, statusbar mode, remappable hold gestures, device lock

Higgs2209:
## Summary

Re-target of #15 against the new `dev` (PlatformIO modular) layout. The original was a single-file `.ino` patch; this version reshapes the same feature set into separate modules that fit the new structure.

### Input layer — `src/hal/input.{cpp,h}`, `src/config.h`, `src/pure/hold_gesture.{cpp,h}`

Two new gesture kinds in the click classifier:

- `ButtonEvent::VeryLong` — hold ≥ `VERY_LONG_MS` (2000 ms) with no preceding click.
- `ButtonEvent::ClickHold` — short click immediately followed by a long hold (a chord). Reuses the existing `clickCount_` machinery; the hold-detection block consumes the pending single when the hold crosses `LONG_MS`.

Plain `Long` no longer fires at the `LONG_MS` hold threshold — at that moment the classifier can't yet tell whether the user is on their way to a very-long. `Long` is now release-time only. `VeryLong` and `ClickHold` still fire at threshold (the gesture is unambiguous by then).

Release-time and in-progress decisions moved into a pure module so the rules can be host-tested without the ISR queue, debounce, or `millis()`:

```cpp
// src/pure/hold_gesture.h
enum HoldGestureKind { HoldNone, HoldLong, HoldVeryLong, HoldClickHold };
HoldGestureKind classifyHoldRelease   (uint32_t dur_ms,  uint8_t click_count);
HoldGestureKind classifyHoldInProgress(uint32_t held_ms, uint8_t click_count);
```

New `Gestures` namespace owns NVS-persisted bindings from the three remappable gestures (Long, VeryLong, ClickHold) to a `ButtonAction` enum (None / Bookmark / Lock / Menu). Defaults: Long → Bookmark (the pre-PR behaviour), VeryLong → Lock, ClickHold → Menu.

`ButtonState::seedPressOnWake()` is called from `setup()` when BTN is held at boot (deep-sleep wake via ext0). The down-edge happened before the ISR was attached and would otherwise be invisible to the classifier; without seeding, "click-then-hold on wake" can't form a chord.

### Reader UI — `src/ui/statusbar.{cpp,h}`, `src/ui/reader_menu.{cpp,h}`, edits to `reader.cpp`, `font.cpp`

Statusbar mode (NVS key `cfg_statusbar`):

- `Full` — 8 px progress bar + page number (the current behaviour).
- `Minimal` — 1 px progress fraction at the very bottom.
- `Hidden` — no statusbar reserve at all.

`Statusbar::reserveH()` is subtracted in `Font::bodyLayout()` so `maxLines` adapts; `Statusbar::setMode()` calls a new public `Font::invalidateLayoutCache()` to force re-pagination.

Reader menu overlay opens via `ACTION_MENU`. Shows book title, page progress, and a statusbar-mode row. Short-click cycles the statusbar mode in place; any other click closes. While open, all other reader input is suppressed.

### Device lock — `src/ui/lock.{cpp,h}`, edits to `Pala_One_2_1.ino`

- `Lock::engage()` sets the flag, persists to NVS (`cfg_locked`); the reader-screen action handler then calls `Sleep::enter()` so the device locks AND sleeps in one motion.
- On wake, if locked, `setup()` keeps the wake-press edges around (calls `markUserActivity` instead of `resetInputFrontend`) so a click-then-hold can wake AND unlock without a second gesture.
- `loop()` swallows all non-unlock input while locked and intentionally does NOT call `markUserActivity` for those events — auto-sleep keeps ticking so an idle locked device re-sleeps rather than draining battery.
- Unlock policy is **permissive**: any of `Long` / `VeryLong` / `ClickHold` unlocks, not strictly the gesture bound to `ACTION_LOCK`. After a deep-sleep wake the firmware can't reliably reconstruct a chord, so strict matching would leave the user stuck if they bound LOCK to the chord.

### Web settings — `src/web/settings.cpp` + `src/lang/{en,es_la}.h`

New "Buttons" card on `/settings` with three `<select>`s for the three remappable gestures. Submitted as a separate form from the Reading section, mirroring the page's existing split.

Strings go through the i18n macro layer introduced in #20:

- `en.h` — 11 new `D_WEB_BUTTONS_*` keys.
- `es_la.h` — 11 mirrors. Translations are best-effort (e.g. "Pulsación larga", "Clic y mantener", "Marcar página"). Tone matches existing entries (formal register, accents on á í ó ú ñ). A native-speaker pass from @tomascorreag or the maintainer would be welcome.

## Tests

`test/test_hold_gesture.cpp` — **11 cases** pinning the gesture rules with concrete `(duration_ms, click_count)` pairs:

Release-time classifier:

- Short release (`dur < LONG_MS`) returns `HoldNone` regardless of `click_count`.
- Solo hold in `[LONG_MS, VERY_LONG_MS)` → `HoldLong`.
- Solo hold `≥ VERY_LONG_MS` → `HoldVeryLong`.
- Pending single + any hold `≥ LONG_MS` → `HoldClickHold` (no upgrade to VeryLong even at 10× — gesture identity is "click then hold", not duration).
- Multi-click (`count ≥ 2`) followed by hold → `HoldNone`. Tap-tap-hold isn't a defined gesture; pending clicks commit on their own and the hold contributes nothing.

In-progress classifier:

- Below thresholds → `HoldNone`.
- `count == 0`, `held ≥ LONG_MS` (but `< VERY_LONG_MS`) → `HoldNone`. Plain Long is deferred to release because at `LONG_MS` we can't yet tell whether the user is on their way to a VeryLong.
- `count == 0`, `held ≥ VERY_LONG_MS` → `HoldVeryLong`.
- `count == 1`, `held ≥ LONG_MS` → `HoldClickHold` (chord crosses threshold).
- Multi-click held → `HoldNone`.

Verified locally on the rebased branch (Windows / MSVC 19.50, all 11 cases + 109 pre-existing tests pass, 0 failures):

```powershell
cmake -S test -B test/build
cmake --build test/build
```

## Scope / history

Four commits:

1. `input: chord gesture, very-long hold, remappable hold actions`
2. `reader: menu overlay, statusbar mode, persistent device lock`
3. `web: Buttons section in settings page for remappable hold actions` (includes the `D_WEB_BUTTONS_*` lang macros)
4. `input: extract hold classifier to src/pure/hold_gesture + host tests`

Rebased on current `dev` (post-PR #20 language selection); auto-merge clean. Independent of the other two feature PRs; if all three land, conflicts are limited to `src/ui/reader.cpp` (each adds a hook call inside `advancePage` / `retreatPage`), `src/ui/sleep.cpp`, `Pala_One_2_1/Pala_One_2_1.ino`, and `test/CMakeLists.txt`. Trivial resolution — keep all calls / all entries.

`Font::invalidateLayoutCache()` is new public API on the Font module — used by `Statusbar::setMode` to force re-pagination after a reserve-height change.


Pala_One_2_1/Pala_One_2_1.ino
```diff
@@ -175,6 +200,36 @@ void loop() {
...
+      Lock::disengage();
+      markUserActivity();
+      Toast::show("Unlocked");
+      // Partial refresh is enough — the underlying page is already on
+      // screen (either from before the lock, or rendered by setup() after
+      // a deep-sleep wake). The render cycle paints the toast.
+      g_currentScreen->draw();
+      return;
+    }
+    if (ENABLE_DEEP_SLEEP && g_currentScreen->allowSleep()) {
```
gnatpat: Nit: it would be nice not to repeat this sleep guard + call - maybe a tiny `deepSleepIfAllowed()` function or something?

Pala_One_2_1/src/ui/font.cpp
```diff
@@ -102,4 +103,6 @@ void setLineGap(int gap) {
 int currentBodySize() { return s_size; }
 int currentLineGap()  { return s_lineGap; }
 
+void invalidateLayoutCache() { s_layoutValid = false; }
```
gnatpat, line 362: I'd have to double check but I don't believe this is enough any more - if this can change the number of lines shown on the screen, we'll need to change the reading cache to know about this as well, as it uses font and line spacing explicitly to work out if the page offset cache is still valid

Pala_One_2_1/src/hal/input.h
```diff
@@ -118,6 +139,41 @@ struct ButtonEvent {
   bool any() const { return kind != None; }
 };
 
+// ============================================================================
+//  Remappable hold-gesture bindings — owned by the input layer because the
+//  bindings are gesture-to-action, not screen-to-screen. The reader screen
+//  asks `actionFor(Kind)` and dispatches to its own action handler.
+// ============================================================================
+
+enum ButtonAction {
```
gnatpat: Question: do you think this input file is the right place for these? IMO they are reader specific and should maybe live closer to the reader rather than the input layer? Just thinking ahead to if another screen someday wants to configure its own actions for each of these presses or something?

Higgs2209:
@gnatpat I think this if finally okay to go. Can you please re-review?

ncont:
I tried it but have the following situation: 

Long Press -> Device goes to Screensaver in Lock Mode. 
Short Press -> Reader opens but any further button interaction is not recognised. 

Long Press shows the "Unlock" text. 

I think that will add confusion, when in Lock mode the pala goes back to reading but does not take any other gesture. 



Another Improvment: Maybe a little overlay text to the Screensaver or a Lock icon could be added. I can imagine that i lock it, forgett it and try to understand why it is not responding. 


Menu works as expected.

Higgs2209:
> I tried it but have the following situation:
> 
> Long Press -> Device goes to Screensaver in Lock Mode. Short Press -> Reader opens but any further button interaction is not recognised.
> 
> Long Press shows the "Unlock" text.
> 
> I think that will add confusion, when in Lock mode the pala goes back to reading but does not take any other gesture.
> 
> Another Improvment: Maybe a little overlay text to the Screensaver or a Lock icon could be added. I can imagine that i lock it, forgett it and try to understand why it is not responding.
> 
> Menu works as expected.

I will admit I used claude to speed this up. Will be able to test in a few hours, else if you want to test feel free to.

Will update the PR once I have had a chance to test it out.

Higgs2209:
@ncont ready to go I believe

PR #28, opened 2026-05-22
Dev server for web UI

Roanmh:
## Summary

- Compiles the real `src/web/` C++ handlers for Linux by stubbing out Arduino-specific dependencies (`WebServer`, `LittleFS`, `Preferences`)
- Uses cpp-httplib v0.15.3 as the HTTP backend, isolated behind a PIMPL boundary to avoid `min`/`max` macro conflicts with the Arduino compat shim
- All six portal pages load: `/`, `/files`, `/bookmarks`, `/list`, `/settings`, `/reset`
- I've tested `/files` the most
- added an integration test stub

## Dev workflow

```bash
# One-time setup
cmake -S dev -B dev/build

# Incremental build (~1–2s with ccache)
cmake --build dev/build
./dev/build/pala_web_emu     # listen on localhost:8080
```

The virtual filesystem is rooted at `dev/fs/` (overridable via `PALA_EMU_FS_ROOT`). A sample book is pre-populated at `dev/fs/books/sample.txt`.

## Test plan

- [x] `cmake -S dev -B dev/build && cmake --build dev/build` compiles clean
- [x] `./dev/build/pala_web_emu` starts and prints listening message
- [x] Smoke test all six pages load in browser at localhost:8080
- [x] `dev/fs/books/sample.txt` appears in the `/files` page
- [x] Edit a string in `src/web/files.cpp`, rebuild, restart — change is visible
- [x] Existing host tests still pass: `cmake -S test -B test/build && cmake --build test/build && ctest --test-dir test/build`
- [x] Integration stub tests pass: `cmake -S dev -B dev/build && cmake --build dev/build && ctest --test-dir dev/build --output-on-failure`

Co-authored by [Claude Code](https://claude.com/claude-code)

PR #29, opened 2026-05-22
Find-in-Book web feature

Roanmh:
## Notes

This PR needs to be moved off of my changes from #28 before it can be merged

I know that Kev1nster's Firmware has a similar feature. We should consider which design we want and whether elements of one design should be applied to another.

## Summary

- **Find-in-Book**: new `/jumptext` POST route lets users search for text in a book from the web UI and jump to the first page containing it. Search is case-insensitive (ASCII fold). The handler saves both the byte offset and page index to NVS so the device opens to the right page on next load. A `?error=not-found` banner is shown when the text isn't present.
- **`src/pure/find_text`**: Boyer-Moore-Horspool `find_text()` for in-memory strings and `findByteOffset()` for chunked `IReadStream` search with overlap to catch cross-chunk matches. Both are host-testable; 10 new unit tests added.
- **`pageOffsetForText`**: walks the page cache (loading and extending the on-disk table) to return the start offset of the page containing the match, rather than the raw byte position.
- **Host web emulator** (`dev/`): native binary serving the same route handlers as the firmware, backed by a local filesystem. Uses the real `paginatePage` with a `helvR12_te` character-width lookup table so page boundaries are close to device output. Includes request/response logging to stderr and a VS Code debug launch config.
- **Integration tests**: 11 end-to-end tests drive the live emulator over HTTP, covering validation, redirect behaviour, offset persistence, and the not-found guard. **DEPENDS on #28**

image

## Test plan

- [x] `cmake -S test -B test/build && cmake --build test/build && ctest --test-dir test/build --output-on-failure` — all unit tests pass
- [x] `cmake -S dev -B dev/build && cmake --build dev/build && ctest --test-dir dev/build --output-on-failure` — all integration tests pass
- [x] Run `dev/build/pala_web_emu`, open `/files`, search for text present and absent in the sample book
- [x] Flash to device and verify the reader opens to the correct page after a web text search

Co-authored by [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Pala_One_2_1/src/pure/arduino_compat.h
+++ b/Pala_One_2_1/src/pure/arduino_compat.h
@@ -57,6 +57,10 @@ public:
   String& operator+=(int n)           { s_ += String(n).s_; return *this; }
   String& operator+=(unsigned long n) { s_ += String(n).s_; return *this; }
 
+  bool concat(const char* buf, unsigned int len) { s_.append(buf, len); return true; }
+  bool concat(const String& o)                   { s_ += o.s_; return true; }
+  bool concat(char c)                            { s_ += c; return true; }
+
   String operator+(const String& o) const { String r(*this); r += o; return r; }
   String operator+(const char* o) const   { String r(*this); r += o; return r; }
   String operator+(char c) const          { String r(*this); r += c; return r; }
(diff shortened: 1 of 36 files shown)


Higgs2209:
I like this implementation - will review in the morning

Roanmh:
This is in a weird state after declining the dev server and I am afk until Wednesday. Feel free to rebase the dev server out if you really want this, otherwise I'll get back to this soon

ncont:
Hey, 

thank you first of all for your contribution!

This feature was merged with #32. It would be very good if you could have a look on the current implementation and compare it to yours. If you find improvments, i would be very happy to check your next PR!

PR #16, opened 2026-05-18
Faster wake up and sleep

PatrykLipka:
please review, try testing all the possible full refreshes to be sure that it works properly

from what I've tested I managed to cut down the 12s wake up into book screen into 4s wake up time

PR #30, opened 2026-05-22
Faster wake-up time

PatrykLipka:
I've opened a new PR, tbh didn't test that yet as a lot of my changes were already introduced in https://github.com/PaulLagier/pala-one-firmware/commit/eaa59ceb3791272bf0a5d0da5513e496fc851ca6 only this during setup is missing which helped with the double clear, I'm just really not sure anymore in which state the code is as it changed a lot since I've opened the first PR: https://github.com/PaulLagier/pala-one-firmware/pull/16


ncont:
That's how t looks on mine after reboot. I think the initial clear seems to be needed.

I tried to use only fast mode and clear but I still get ghosting.

ncont:
Based on that and that most of your changes found already the way into the code - i think we can close the pr.

PatrykLipka:
> Based on that and that most of your changes found already the way into the code - i think we can close the pr.

ah, okay, I will change the code then, we'd need to add the slow mode to menu loading again as I did in my previous PR (before the codebase changed)

I'll try to figure it out in a couple of days, as with change from this PR I believe book will load 2x faster (4s instead of 8s)

ncont:
Sounds good! Could you put this pr into draft and close the old one?

PR #31, opened 2026-05-23
Dark mode for the web UI

Chris24680:
Ports the dark-mode WebUI from @kevinst1r's fork (https://github.com/kevinst1r/pala-one-firmware-kevinst1r) onto the modular `src/` layout. Original feature design and implementation by @kevinst1r.

## Summary

Rewrites `src/web/chrome.cpp` so every color in `/style.css` flows through CSS custom properties:
- `:root` carries the light palette (the existing values, unchanged).
- `html[data-theme=dark]` overrides with the dark palette.

A pre-paint inline `<script>` is injected into `<head>` by `webPageStart()` — it reads `localStorage.palaTheme`, falls back to `prefers-color-scheme`, and sets `document.documentElement.dataset.theme` before first paint so there's no flash of the wrong palette. A small toggle button (`Dark mode` / `Light mode`) is added to the top-side nav row.

Per-browser preference only — no NVS persistence, no new firmware state.

## Test plan

- [x] Build clean on `wireless-paper-v1_2` (default `LANG_EN`) — already verified locally.
- [ ] Open the web UI on a desktop browser; click the toggle. Theme flips immediately; reload preserves it.
- [ ] Open the web UI on a phone with system dark mode on. First load picks dark automatically. Toggle to light; reload stays light.
- [ ] Visit `/`, `/files`, `/settings`, `/bookmarks` — all render correctly under both themes.

## Notes

- Only `src/web/chrome.cpp` changes.
- No new `lang/` keys needed (toggle label text is purely CSS-driven via `.tgl` / `.tgd` siblings).

### DIFF
--- a/Pala_One_2_1/Pala_One_2_1.ino
+++ b/Pala_One_2_1/Pala_One_2_1.ino
@@ -36,6 +36,16 @@
 // #define LANG_ES_LA
 // ────────────────────────────────────────────────────────────────────────────
 
+// ── Web UI default theme: uncomment exactly one ─────────────────────────────
+//   The web UI has a light and a dark palette and a per-page toggle button.
+//   Once a visitor picks one the choice is remembered in their browser's
+//   localStorage and overrides whatever's set here — this define only picks
+//   the *first-visit* default. Default if nothing is set: light.
+//   PlatformIO users can also pass -D WEB_THEME_DARK in build_flags.
+#define WEB_THEME_LIGHT
+// #define WEB_THEME_DARK
+// ────────────────────────────────────────────────────────────────────────────
+
 // When built with PlatformIO, WIRELESS_PAPER + DISPLAY_V1_x come from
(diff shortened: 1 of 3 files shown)


ncont:
Tested it and it works!

But i would have one suggestion: I would leave the default Theme "light". To be able to change it permanently, a define in the ino would be nice - same as the language. 

And a section in the readme for the user would be good. :)

Higgs2209:
> Tested it and it works!
> 
> But i would have one suggestion: I would leave the default Theme "light". To be able to change it permanently, a define in the ino would be nice - same as the language.
> 
> And a section in the readme for the user would be good. :)

Agree - default theme should definitely be light mode and a firmware toggle would be nice also.

Chris24680:
Done!

Higgs2209:
Testing now

ncont (approved):

Higgs2209 (approved):
Code looks good and works on device as expected

PR #32, opened 2026-05-23
In-browser book reader with find and jump-to-offset

Chris24680:
Ports the in-browser reader + find/jump from @kevinst1r's fork (https://github.com/kevinst1r/pala-one-firmware-kevinst1r) onto the modular `src/` layout. Original feature design and implementation by @kevinst1r.

## Summary

New module `src/web/find.{h,cpp}` adds three routes:
- `GET /read?id=N` — in-browser reader with find UI (search field, prev/next/jump buttons, page jump form).
- `GET /readbook-text?id=N` — streams the raw normalized book text; browser does the search client-side.
- `POST /jumpoffset` — sets the resume byte offset for book `id`. If the reader is currently open on that book, also updates the in-memory cursor so the device sees the jump on the next render.

Per-book "Read & find" link added to the existing `/files` page. Page-number jump uses the existing `POST /jumppage` route.

`D_WEB_READ_*` macros added to `src/lang/en.h` and `src/lang/es_la.h`. JS-internal status strings (`"No matches"`, `"Match N of M"`, etc.) are left as English with a TODO for a future i18n pass.

## Test plan

- [ ] Build clean on `wireless-paper-v1_2` for both `LANG_EN` and `LANG_ES_LA` — already verified locally.
- [ ] Open `/files`, click `Read & find` on a book → `/read?id=N` loads, text streams in, status shows byte count.
- [ ] Type a phrase, press Enter (or Find all). Matches highlight; status shows `Match 1 of N`. Next / Prev cycle. Click a highlighted hit → that match becomes active.
- [ ] Click `Jump to here` → status confirms save. Open the book on the device → reader lands at the corresponding page.
- [ ] Enter a page number, click `Jump to page` → device reader resumes at that page on next open.
- [ ] Open the same book in `/read` while the reader is already running it on-device → after `/jumpoffset` the device cursor updates without a reopen.

## Notes

- Self-contained: no changes to other features. The existing `/jumppage` from `src/web/files.cpp` is reused; `/jumpoffset` is the new endpoint.


ncont (approved):
Works for me.

PR #34, opened 2026-05-23
Bionic reading, OpenDyslexic font, and reading-location retention

Chris24680:
Ports bionic reading + OpenDyslexic font + reading-location retention from @kevinst1r's fork (https://github.com/kevinst1r/pala-one-firmware-kevinst1r) onto the modular `src/` layout. Original feature design and implementation by @kevinst1r.

### `src/ui/font.{h,cpp}` + `Pala_One_2_1/opendyslexic_u8g2_fonts.h`
- `enum class Family { Helvetica, OpenDyslexic }` — chosen at runtime via NVS `cfg_font_fam`. OpenDyslexic tables vendored as `opendyslexic_u8g2_fonts.h` at the sketch root; only referenced from `font.cpp`.
- `setFamily` / `setBionic` / `currentFamily` / `bionicEnabled` accessors. NVS keys: `cfg_font_fam`, `cfg_bionic`.
- `measureBionicLine` / `drawBionicLine` — UTF-8-aware word splitting with a 1px gap between the bold prefix and the regular tail. Same "strong char count" rule as @kevinst1r's version (≤3 chars → 1; otherwise `chars/2` clamped to [1, 4]). When bionic is off, both collapse to a plain `u8g2.getUTF8Width` / `u8g2.print` under the Body font.

### `src/ui/text.cpp`
- `bodyMeasure` (paginator's `MeasureFn`) delegates to `Font::measureBionicLine`, so wrap budget stays correct when bionic is on.
- `drawPageAt`'s `onLine` callback uses `Font::drawBionicLine`.

### `src/storage/page_cache.{h,cpp}` (file format bump)
- New `PageCacheLayout { bodySize, lineGap, family, bionic }` struct replaces the two-int stamp.
- `encodeLayoutVersion` packs four bytes into a 32-bit field (was 16-bit `(size << 8) | gap`).
- Magic bumped `0x50434F47 → 0x50434F48` so older caches are silently rejected and rebuilt on the next save. No migration code needed.

`src/ui/reader.cpp` updated to pass `Font::layoutForCache()` at the two call sites.

### `src/web/settings.cpp` — UI + position retention
- New form fields: font family dropdown (Helvetica / OpenDyslexic), bionic reading checkbox, intro text explaining position retention.
- `applySettingsForm()` returns `bool layoutChanged` covering size / family / lineGap / bionic.
- `handleSettingsPost` snapshots `g_bookview.pages.offsets[cursor.pageIndex]` (the current top-of-page byte) before applying changes. After applying, it resets the in-memory page table, calls `findPageForOffset(savedByte)` to land on the page that contains that byte under the new layout, persists the new page index, and re-renders if the reader is active. The on-disk page cache self-invalidates via the layout stamp — no explicit cache wipe needed.

`D_WEB_FONT_FAMILY_*` / `D_WEB_BIONIC_*` / `D_WEB_READING_INTRO` / `D_WEB_SETTINGS_APPLY_HINT` macros added to `src/lang/en.h` and `src/lang/es_la.h`.

## Test plan

- [ ] Build clean on `wireless-paper-v1_2` for both `LANG_EN` and `LANG_ES_LA` — already verified locally.
- [ ] Open a book on the device. Cycle through all four font sizes (8/10/12/14) × both families (Helvetica / OpenDyslexic) × both line gaps via `/settings`. Text reflows cleanly, no right-edge overflow, no missing glyphs.
- [ ] Toggle bionic — first ~40% of each word renders bold with a 1px gap to the regular tail. Position retained across the toggle.
- [ ] Open a book at ~page 50, change font size from 10 → 14 via `/settings`, return to device → lands within ±1 page of the same byte offset.
- [ ] Existing on-disk `pc_*.bin` caches (magic `0x50434F47`) are silently rejected and rebuilt on first open — no error log spam, no data loss.

## Notes

- Position retention is part of this PR because the `layoutChanged` flag tracks the font family + bionic state, which only exist with this change. Splitting them further would require a layered branch.
- The page-cache magic bump means *all* existing per-book caches will be one-time invalidated on the device's first run with this firmware. Books re-paginate transparently; users will see a brief delay on first page turn per book.


ncont (approved):
Looks good to me. Resolve conflicts and then it is ready to merge.

PR #35, opened 2026-05-23
Add Samim 9pt font option and no-screensaver mode

quintenluyten2:
## Samim 9px font (size 9)

Adds `u8g2_font_samim_12_t_all` as a new font size option in Settings → Reading.

**Known shortcomings — please read before merging:**

- **Special characters in Latin text** Samim is a Persian/Arabic script font. Selecting size 9 on a device loaded with Latin-alphabet books may cause issues for languages such as French due to limited support for special characters in the Latin alphabet. However, Samim 9px is still very much suitable for English readers.
- **No bold variant.** U8g2 does not include a bold face for Samim. `s_bold` is aliased to the same `samim_12_t_all` pointer as `s_body`, so any bold-emphasis call (e.g. section headers, selected menu rows) renders identically to the regular weight.
- **Different encoding suffix.** All other body sizes use `_te` (extended Latin). Samim uses `_t_all` (all Unicode, transparent background). Code paths that assume `_te` coverage — particularly toast messages with accented characters from translations — may render missing-glyph boxes when size 9 is active.
- **Ascent correction.** Samim 12 reports an ascent 1 px taller than its visual cap height warrants relative to the Helvetica faces. Both `bodyLayout()` (reader) and `menuLineH()` (menus) subtract 1 when size == 9 to keep line counts comparable.

The `applyBodySize()` default falls back to size 10 for any unrecognised value, so no existing user is affected.

---

## No-screensaver mode

Adds a checkbox toggle in Settings → Reading (NVS key `cfg_noscr`, default `false`). When enabled:

- **Sleep still happens normally** on the usual idle timer.
- **`Sleep::enter()` skips `drawSleepScreen()`** — the last page stays visible on the e-ink panel instead of being replaced by the screensaver image.
- **Wake skips the full display refresh.** `setup()` now reads `cfg_noscr` before `display.clear()` (requires moving `prefs.begin()` + `Sleep::loadSettings()` a few lines earlier in setup). On a deep-sleep wake with no-screensaver active, `display.clear()` is skipped so the panel is not blanked before the fast-mode redraw of the restored page.

Changes are confined to `sleep.cpp/h`, `web/settings.cpp`, the two language headers, and a small reorder in `Pala_One_2_1.ino`.

---

## Files changed

| File | What changed |
|------|-------------|
| `src/ui/font.cpp` | Add `case 9` (Samim) in `applyBodySize()`; subtract 1 from ascent in `bodyLayout()` when size == 9 |
| `src/ui/widgets.cpp` | Same −1 ascent correction in `menuLineH()` |
| `src/ui/sleep.cpp` | Add `s_noScreensaver` flag, `loadSettings()`, `setNoScreensaver()`, `noScreensaver()`; gate `drawSleepScreen()` on the flag |
| `src/ui/sleep.h` | Expose `noScreensaver()` / `setNoScreensaver()` |
| `src/web/settings.cpp` | Add size-9 option; add no-screensaver checkbox + POST handler |
| `src/lang/en.h` | `D_WEB_FONT_SIZE_9`, `D_WEB_NO_SCREENSAVER_LABEL/HINT` |
| `src/lang/es_la.h` | Same strings in Spanish |
| `Pala_One_2_1.ino` | Move `prefs.begin()` + `Sleep::loadSettings()` before `display.clear()`; gate `display.clear()` on wake cause + no-screensaver flag |


Higgs2209:
@quintenluyten2 Are you able to please split this up into to separate PRs? They aren't very related (as far as I can see) and would be better dealt with independently.  Thanks :)

quintenluyten2:
Splitting into two separate PRs — one for Samim 9pt font, one for no-screensaver mode.

PR #36, opened 2026-05-23
Add Samim 9pt font size option

quintenluyten2:
Adds `u8g2_font_samim_12_t_all` as a new **9px** size option in Settings → Reading. Samim renders both Latin/English and Persian/Arabic text correctly. The main benefit is an extra option inbetween 8px, which may be too small for some people, and 10px, which shows way fewer words than 8px.

## Known limitations

- **No bold variant.** U8g2 does not include a bold face for this typeface. `s_bold` is aliased to the same `samim_12_t_all` pointer, so bold-emphasis calls (selected menu rows, section headers) render at regular weight.
- **Different encoding suffix.** All other body sizes use `_te` (extended Latin). Samim uses `_t_all` (all Unicode, transparent background). Built-in UI strings are all basic ASCII so this has no practical effect, but it's worth noting if anyone adds toast strings with characters outside `_t_all` coverage.
- **Ascent correction.** Samim 12 reports an ascent 1 px taller than its visual cap height relative to the Helvetica faces. Both `bodyLayout()` (reader) and `menuLineH()` (menus) subtract 1 when size == 9 so a comparable number of lines fit on screen.

The `applyBodySize()` default falls back to size 10 for any unrecognised value — no existing user is affected.

## Files changed

| File | What changed |
|------|-------------|
| `src/ui/font.cpp` | Add `case 9` (Samim) in `applyBodySize()`; −1 ascent correction in `bodyLayout()` when size == 9 |
| `src/ui/widgets.cpp` | Same −1 ascent correction in `menuLineH()` |
| `src/web/settings.cpp` | Add size-9 option in the font-size `<select>` |
| `src/lang/en.h` | `D_WEB_FONT_SIZE_9 "9px — Samim"` |
| `src/lang/es_la.h` | Same string in Spanish |

### DIFF
--- a/Pala_One_2_1/src/lang/en.h
+++ b/Pala_One_2_1/src/lang/en.h
@@ -242,6 +242,7 @@
 #define D_WEB_READING_HEADING       "Reading"
 #define D_WEB_FONT_SIZE_LABEL       "Font size"
 #define D_WEB_FONT_SIZE_8           "8px &mdash; tiny"
+#define D_WEB_FONT_SIZE_9           "9px &mdash; Samim"
 #define D_WEB_FONT_SIZE_10          "10px &mdash; small"
 #define D_WEB_FONT_SIZE_12          "12px &mdash; medium"
 #define D_WEB_FONT_SIZE_14          "14px &mdash; large"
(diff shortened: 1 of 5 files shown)


quintenluyten2:
I found some issues with text rendering that need to be solved before this can be merged

ncont:
Please reopen when you updated. I will close this stale PR for now.

PR #37, opened 2026-05-23
Add no-screensaver mode

quintenluyten2:
Adds a **No-screensaver mode** checkbox in Settings → Reading (NVS key \`cfg_noscr\`, default \`false\`).

## Behaviour when enabled

- The device still goes to sleep on the normal idle timer — nothing changes there.
- **Only applies when reading a book.** If the user falls asleep in the library or a menu, the screensaver is shown as normal.
- **When going to sleep from the reader:** the screensaver image is skipped. Instead, \`display.fastmodeOff() + display.update()\` flushes the reader's framebuffer with a full refresh, so the last page sits cleanly on the panel before power-down.
- **On wake from deep sleep:** \`setup()\` skips \`display.clear()\` so the panel is not blanked before the fast-mode redraw of the restored page. The user can continue reading with a single button click, without the interruption of a full display refresh.

The wake-side skip is gated on all three: woke-from-sleep AND no-screensaver enabled AND \`wake_path\` is set (i.e. we were reading). \`wake_path\` is written by \`armResumeOnWake()\` inside \`ReaderScreen::onSleep()\`, which runs before the sleep-image decision — so it doubles as a zero-coupling "was in reader" signal in both \`Sleep::enter()\` and \`setup()\`.

## Implementation notes

- Hidden \`noscr_form\` sentinel in the settings form body: lets \`handleSettingsPost()\` distinguish an explicit unchecked checkbox from an unrelated POST.
- \`Sleep::noScreensaver()\` / \`Sleep::setNoScreensaver()\` follow the same ownership pattern as the existing \`idleTimeoutSecs\` / \`setIdleTimeout\` pair — setting + NVS key are file-private to \`sleep.cpp\`.

## Files changed

| File | What changed |
|------|-------------|
| \`src/ui/sleep.cpp\` | \`s_noScreensaver\` flag; in \`enter()\`, checks \`wake_path\` to skip screensaver only from reader; does full refresh with page content instead |
| \`src/ui/sleep.h\` | Expose \`noScreensaver()\` / \`setNoScreensaver()\` |
| \`src/web/settings.cpp\` | No-screensaver checkbox + POST sentinel handling |
| \`src/lang/en.h\` | \`D_WEB_NO_SCREENSAVER_LABEL\` / \`D_WEB_NO_SCREENSAVER_HINT\` |
| \`src/lang/es_la.h\` | Same strings in Spanish |
| \`Pala_One_2_1.ino\` | Move \`prefs.begin()\` + \`Sleep::loadSettings()\` before \`display.clear()\`; gate \`display.clear()\` on wake cause + flag + \`wake_path\` |

### DIFF
--- a/Pala_One_2_1/Pala_One_2_1.ino
+++ b/Pala_One_2_1/Pala_One_2_1.ino
@@ -133,12 +133,17 @@ void setup() {
   prefs.begin("ereader", false);
   Sleep::loadSettings();
 
-  // In no-screensaver mode, skip the full-refresh clear on deep-sleep wake so
-  // the last page is not blanked before the fast-mode redraw. On a fresh boot
-  // (or normal screensaver mode) always clear to white first.
+  // Skip the full-refresh boot clear only when all three are true:
+  //   (a) waking from deep sleep (not a fresh boot),
+  //   (b) no-screensaver mode is on, and
+  //   (c) we were reading a book (wake_path was set by armResumeOnWake during
+  //       the sleep entry — tryRestoreReadingSession() will consume it later).
+  // If the user fell asleep in a menu the screensaver was shown, so we always
+  // clear normally in that case.
   bool wokeFromSleep = (esp_sleep_get_wakeup_cause() == ESP_SLEEP_WAKEUP_EXT0);
+  bool wereReading   = (prefs.getString("wake_path", "").length() > 0);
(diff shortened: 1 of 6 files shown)


Higgs2209 (commented):
Happy with this - Tested on device and it compiles and works as expected. Useful change and is clear in the webUI.

PR #39, opened 2026-05-23
 Windows PowerShell builder for example apps

Higgs2209:
examples/build.ps1: native PowerShell builder for the example apps. Uses the xtensa toolchain that Arduino IDE installs at %LOCALAPPDATA%\Arduino15\packages\esp32\tools\esp-x32\ plus the system Python. No make or Git Bash needed.

The Makefile path stays the default for Linux / macOS users; this is a parallel option for Windows users who don't want WSL or a separate make install. Both produce identical .bin output (same compile + objcopy + Python patch sequence).

### DIFF
--- a/examples/build.ps1
+++ b/examples/build.ps1
@@ -1,0 +1,122 @@
+# Pala One app builder for Windows.
+#
+# Builds a single example app without requiring `make` or Git Bash. Uses the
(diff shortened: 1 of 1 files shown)


ncont:
I have no windows to test. But since it can not brake anything, it should be good.

PR #44, opened 2026-05-24
Remove LoRa radio features to avoid board damage

Chris24680:
## Summary

- Removes `lora_driver.cpp` / `lora_driver.h` (SX1262 RadioLib driver)
- Removes the `mesh_chat` example app
- Removes the 6 LoRa function pointers from `PalaAPI` (`loraInit`, `loraSleep`, `loraReady`, `loraRecv`, `loraSend`, `loraNodeId`)
- Removes the `RadioLib` dependency from `platformio.ini`
- Removes the `loraSleep()` call from `sleep.cpp`

## Why

The Heltec Wireless Paper uses an SX1262 LoRa radio. **Transmitting without an antenna attached can permanently damage the radio chip** — even a few seconds of transmission is enough to cause irreversible harm ([nodakmesh.org/setup/wireless-paper](https://nodakmesh.org/setup/wireless-paper)).

The firmware as shipped on `dev` initialises the LoRa radio and exposes it to loadable apps via `PalaAPI`. There is no hardware interlock or software guard preventing transmission when no antenna is connected. Any app (or future app) calling `loraInit` followed by `loraSend` will immediately put the chip at risk.

Removing the radio driver and API surface eliminates that risk entirely. The `PalaAPI` struct remains at v4 and all other functionality is unchanged — this branch is otherwise identical to `origin/dev`.

## Future work

In the future it would be worth looking at safely reintroducing LoRa support, perhaps as an opt-in build flag (e.g. `-D ENABLE_LORA`) so it is off by default and users must consciously enable it — ideally accompanied by a clear warning in the build docs that an antenna must be attached before flashing.

## Test plan

- [x] Build `wireless-paper-v1_2-en` and confirm no `RadioLib` or `lora_driver` symbols appear
- [x] Confirm `grep -r "lora\|RadioLib" Pala_One_2_1/` returns no results
- [x] Flash and verify normal reader/app functionality is unaffected


gnatpat (approved):
Given this was never merged to main, I reckon we could bump the API version back down to 3? Keeps more apps compatiable? Up to you though

Chris24680:
> Given this was never merged to main, I reckon we could bump the API version back down to 3? Keeps more apps compatiable? Up to you though

I'm happy with either, I wasn't sure if keeping it as 3 or 4 would be the most compatible, but I guess you're right

Higgs2209 (approved):

PR #45, opened 2026-05-24
Firmware Statistics screen

Higgs2209:
Replaces what would have been two separate apps (reading_streak + stats) with a single firmware-internal Statistics screen, addressing reviewer feedback on PRs #25 / #26: apps shouldn't share wire-format structs with the firmware via files in /apps/<key>.dat; that contract is supposed to be PalaAPI-only. Moving the views into firmware eliminates the shared-struct coupling and the file format entirely.

New modules:

  - src/storage/statistics.{cpp,h} — combined state owner. Lifetime counters (pagesRead, buttonPresses) live in RTC RAM (preserved across deep sleep, zero flash writes during normal use) and flush to NVS every 100 events or on Sleep::enter. Streak state is the existing ReadingStreakFile layout from src/pure/streak_log.h, persisted straight to NVS under cfg_streak (log events are at most once per real day, so flash wear isn't a concern). The bitmap-shift + continuation rules still live in src/pure/streak_log.cpp — host-tested.

  - src/ui/screens/statistics_screen.{cpp,h} — view-only Screen. Shows current/longest streak, total sessions, lifetime pages, lifetime button-presses, and a grid bitmape showing which days hit the streak threshold. Any click returns to the library.

Wiring:

  - LIB_ENTRY_STATISTICS added to LibraryEntryType, slotted between Apps and Device in the library menu.
  - Statistics::onReaderPageTurn() in advancePage / retreatPage.
  - Statistics::bumpButtons(delta) from loop() against a diff of g_btns.peekPressCount() (new const accessor on input.h that doesn't disturb the apps-API consumer).
  - Statistics::loadOnBoot() in setup() — reload counters from NVS.
  - Statistics::flushToNvs() in Sleep::enter() — drain deltas.

Lang (i18n via the #20 macro layer):

  - D_MENU_STATISTICS + 4 D_STATS_* snprintf templates added to src/lang/en.h and src/lang/es_la.h. Spanish translations are best-effort; a native-speaker pass is welcome.

Tests:

  - src/pure/streak_log.{cpp,h} carried over unchanged from the earlier port — pure log-advance rules.
  - test/test_streak_log.cpp — 7 host-test cases pinning the bitmap-shift, continuation, gap-reset, and STREAK_DAY_UNSET semantics. Verified locally (104 tests / 306 checks / 0 fail on Windows).

What's NOT here:

  - The reading_streak and stats example apps — removed (no longer needed; firmware owns the data).
  - The sleep_timer app — split to its own branch (dev/sleep-timer) to address the  "what does it do?" question separately. Will wait untill after v3 has shipped to open this PR. 
  - examples/build.ps1 — split to its own branch (dev/win-app-builder) #39 

# Review Checklist

- [x] Build compiles and flashes
- [x] Library shows a **Statistics** entry between **Apps** and **Device**.
- [x] Double-click → Statistics screen renders.
- [x] Current streak shows 0
- [x] read 5 pages of a book - check that statistics update


ncont:
Tested: Build on Debian with PlatformIO

- Statistics Menu exist
- Moved around in menu and checked "butten pressed": Count as expected
- Checked Single, Double and Triple Click: Counted all accordingly
- Tested Page Turn: As expected
- Tested Page Turn and Reverse: Counts as expected

Issue #49, opened 2026-05-25
Make Header Title configurable


ncont:
User requesed to be able to change the Menu Title: 
`
  #define LIB_HEADER_TITLE "Pala One " BUILD_GIT_HASH
`

tomascorreag:
I'll add the feature

Issue #52, opened 2026-05-25
'unknown' when no custom title set


kevinst1r:
'unknown' is displayed when no custom title is set.  It should just say Pala One with nothing after it with no custom title

kevinst1r:
Even with one set, it still says 'unknown' after

kevinst1r:
<img width="1183" height="2560" alt="Image" src="https://github.com/user-attachments/assets/c45e1dce-0d0f-4792-a1be-e0d95bc83f16" />

kevinst1r:
UPDATE:  Editing the custom name in WebUI should be added.

tomascorreag:
I agree we should be able to rename the books.

ncont:
The behavior is correct. If a detached src code is delayed, the version is "unknown". 

The name can now be set in the WebUI. Making any adaption obsolete.

PR #56, opened 2026-05-25
settings: user-configurable library header title

tomascorreag:
## Summary

Adds a "Device" card to the web settings page (`/settings`) that lets users set the library screen header to any string — including empty to hide it entirely. Persisted via NVS key `cfg_hdr_title`; defaults to the compile-time `LIB_HEADER_TITLE` when unset.

Closes issue #49

## Changes

- New `src/ui/header_title.{h,cpp}` module (mirrors `Sleep`/`Font` pattern): `loadSettings()` at boot, `current()` for the display, `set()`/`resetToDefault()` for the web handler
- Library screen reads `HeaderTitle::current()` instead of the `LIB_HEADER_TITLE` macro
- Web settings page gets a text input (max 31 chars) with a "Reset to default" checkbox, in its own form card so it doesn't trigger the reader layout-remap logic
- HTML attribute escaping on the rendered value to prevent XSS
- Both `en.h` and `es_la.h` lang strings added

## Notes

- `prefs.isKey()` distinguishes "never set" (use compile-time default) from "set to empty string" (show no title)
- 31-byte cap is a practical limit — bold Helvetica on a 250px-wide display fits ~22 chars after margins + battery indicator. Overflow clips at screen edge.
- Factory reset (`prefs.clear()`) naturally restores the default since the NVS key is removed

## Test plan

- [x] `pio run` all 4 leaf envs (v1_1-en, v1_1-es, v1_2-en, v1_2-es)
- [x] Host CMake tests pass (no `config.h` contamination)
- [x] Flash, enter upload mode, open `/settings`, set custom title → confirm library header changes
- [x] Set empty string → confirm header line disappears
- [x] Check "Reset to default" → confirm original title restored

### DIFF
--- a/Pala_One_2_1/Pala_One_2_1.ino
+++ b/Pala_One_2_1/Pala_One_2_1.ino
@@ -102,6 +102,7 @@
 #include "src/ui/screens/reader_screen.h"
 #include "src/ui/screens/statistics_screen.h"
 #include "src/ui/screens/upload_screen.h"
+#include "src/ui/header_title.h"
 #include "src/ui/lock.h"
 #include "src/ui/screensavers.h"
 #include "src/ui/sleep.h"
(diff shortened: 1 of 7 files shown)


ncont:
Works!

Since it also closes #52, we include it in v3.0 already

PR #58, opened 2026-05-25
Tests and default fallback for language translation items

tomascorreag:
## Summary

- Add host test that verifies all language files define the same `D_*` keys as `en.h` and that no key has an empty or whitespace-only value
- Change `lang.h` to include `en.h` first as a baseline, then overlay the target language — missing translations silently fall back to English instead of producing a compile error
- Add `#undef` before each `#define` in `es_la.h` so the overlay works on all compilers (GCC, Clang, MSVC) without macro-redefinition warnings

### Why

Previously, a missing key in `es_la.h` caused a compile error on the Spanish build. This is safe but blocks incremental translation work — every new key in `en.h` must be translated before the Spanish build compiles. Developers building only `LANG_EN` wouldn't notice the break until much later.

Now the Spanish build always succeeds (falling back to English for untranslated keys), and the host test suite catches drift immediately with specific key names.

### Changed files

| File | What |
|---|---|
| `Pala_One_2_1/src/lang/lang.h` | Include `en.h` first as baseline, then target language as overlay. Updated authoring rules. |
| `Pala_One_2_1/src/lang/es_la.h` | Added `#undef D_*` before each `#define D_*` (337 keys, mechanical). |
| `test/test_lang_keys.cpp` | New. Scans `LANG_DIR` for all `.h` language files, checks key parity against `en.h`, checks for empty/whitespace-only values. |
| `test/lang_config.h.in` | New. CMake `configure_file` template that injects the lang directory path into the test binary. |
| `test/CMakeLists.txt` | Register `test_lang_keys.cpp`, generate `lang_config.h` via `configure_file`. |

## Test plan

- [x] `ctest` passes with all keys in sync (117 tests, 0 failures)
- [x] Removing a key from `es_la.h` → build succeeds, test fails and names the missing key
- [x] Adding a key to `en.h` only → build succeeds, test fails and names the missing key
- [x] Adding a third language file to `src/lang/` would be automatically checked by the test (scans directory)


tomascorreag:
I'm not very happy with the huge amount of #undef declarations in alternate language files, but I couldn't find a way around it while keeping the English fallback. Open to suggestions!

tomascorreag:
@ncont I'm thinking I could make it a simpler json the translator contibutor would fill out, paired with a python script to generate the .h file from that.

ncont:
I am not sure yet. Even thought it does not look nice, it is easy to understand. Especially, since there are quite a lot of new programmers in the project. 

If we would introduce this abstraction layer, it is harder to maintain and understand. 

I'll need to sleep a night over it or maybe someone else has a opinion about it.

tomascorreag:
sure, ty! There's no rush to merge this.

Higgs2209:
@tomascorreag - I agree with @ncont, I think it is easy to understand which I believe is very important for a project of this kind.

Can you resolve the conflicts?

ncont:
Sorry. That I left this pr hanging. 

Tbh, i don't think the current complexity is worth it. If there is anywhere in the Generation Script an issue with the reg exp or so, it takes quite some time to understand. 

Before I was able to hover over the #define and my ide showed me the text. But with the generation script in between it is not able to. 

I would actually prefer rather the solution with just define or the added #ifundef.

tomascorreag:
@ncont so your saying I should revert the .json + .py changes and just keep it with a bunch of #define and #ifundef and such?

ncont:
> @ncont so your saying I should revert the .json + .py changes and just keep it with a bunch of #define and #ifundef and such?

I would say yes. If at all. 

As I understand, the idea is to provide earlier feedback, if a language def is not defined.

All packages are build in CI. So an error would be catched there. Than it is just one simple fix an done. 

In case there is somewhere a logic error in the 1.2k lines of python code, it will take much much more time to understand and debug. 

So, even though the idea is nice and it most probably does work, I think it introduces way to much complexity for a not so critical issue. 

I would rather tend to even don't write all the ifundev. But leave the defines and add a pre commit hook in combination with `pio check ` to statically test all defines. 

But that's my very own opinion. If @Higgs2209 likes your idea, I don't want to block it.

tomascorreag:
I agree that's a better approach for this situation. Will make the changes!

PR #60, opened 2026-05-25
Fix: Screen Lock

### LINKED ISSUE #48: PalaOne does not exit Lock Screen
- Flashed 9b5194bd054da30a6303c7b498afa03fcd972a8c on the device. 
- Entered Lock Screen
- Used long button press to ender Lock screen
- Lock Symbol appears

- Long Press to unlock
- Screen blicks but does not unlock. 

Intersingly, after a bit of time, the lock screen did evetually unlock. Not related to a button press tho.

### LINKED ISSUE #92: locking doesn't work in the main menu
**hardware:** epaper 1.2
**firmware:** release v3.0

**problem:** I have extra long press configured to lock the device. This may be by design, but locking only works when in a book. I think it would be better if you could also lock from the main menu.

ncont:
### Problem

Three bugs combined to make the lock screen nearly impossible to exit with a Long press:

**1. `seedPressOnWake(millis())` underestimates hold duration**

`setup()` runs `delay(200)` before attaching the button ISR. By the time
`seedPressOnWake` is called, `millis()` is already ~202 ms. This value was stored
as `pressStart_`, so the measured hold duration was `release_time − 202 ms`.
A user holding for 900 ms total only got 698 ms counted — below `LONG_MS = 850 ms`
— so the gesture was misclassified as a Short click and silently absorbed as the
"wake press", making it look like nothing happened.

**2. Locked-idle timeout fires while button is held**

The 1500 ms re-sleep timer ran without checking whether the button was currently
pressed. If the timer expired mid-hold, the device entered deep sleep, the
Long-press release happened with no ISR running, and the gesture was lost entirely.
On the next ext0 wake the cycle repeated, causing the screen to blink (full refresh
from `drawSleepScreen`) while staying locked.

**3. Reader overrides `fastmodeOff()` before `display.update()`**

`renderCurrentPage()` always calls `display.fastmodeOn()` or `display.fastmodeOff()`
based on its own `pageTurnsSinceFull` counter (`FULL_REFRESH_EVERY_N_PAGES = 100`).
The unlock path set `display.fastmodeOff()` before `g_currentScreen->draw()`, but
the reader immediately overrode it with `fastmodeOn()` — so `display.update()` ran
in fast mode and screensaver ghosting remained visible, making the device appear
still locked.

### Fix

| File | Change |
|---|---|
| `Pala_One_2_1.ino` | `seedPressOnWake(0)` — measures hold from boot-start, giving full credit for time held before the MCU started |
| `Pala_One_2_1.ino` | `&& !g_btns.isPressed()` guard on the locked-idle sleep gate — mirrors the same pattern already used in `idleLightSleep()` |
| `reader.cpp` / `reader.h` | `forceNextRenderFull()` — one-shot flag consumed by `renderCurrentPage()`, forcing `fastmodeOff` regardless of `pageTurnsSinceFull` |
| `Pala_One_2_1.ino` | Unlock path calls `forceNextRenderFull()` before `g_currentScreen->draw()` |

### Testing

- [x] Tested on V1.1 / V1.2
- [x] Long press, VeryLong press, and ClickHold all unlock correctly
- [x] No ghosting remains after unlock from screensaver
- [x] Normal page-turn fast mode is unaffected (`s_forceNextFull` is one-shot)

### DIFF
--- a/Pala_One_2_1/Pala_One_2_1.ino
+++ b/Pala_One_2_1/Pala_One_2_1.ino
@@ -143,8 +143,13 @@ void setup() {
   // never made it into the queue. Seed the input state so the upcoming
   // release edge is classified as a real press, not silently dropped.
   // Required for "click-then-hold on wake" to form a single chord gesture.
+  // Pass 0 (not millis()) so the duration is measured from boot-start, giving
+  // the full boot time as credit toward the hold — the button press triggered
+  // the ext0 wakeup, so it was down before the MCU started. Using millis() here
+  // would subtract the ~200ms delay() above and incorrectly shorten the
+  // measured hold, causing a Long press to be misclassified as Short.
   if (digitalRead(BTN) == LOW) {
-    g_btns.seedPressOnWake(millis());
+    g_btns.seedPressOnWake(0);
   }
 
   u8g2.begin(gfx);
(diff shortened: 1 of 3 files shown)


Higgs2209 (approved):
Unable to flash and test on device.

Looks like the fix will function based on your edits. 

Thanks

Issue #71, opened 2026-05-28
Book menu issues


PatrykLipka:
As of 929191d menu that you can go into in book view (by default short press + hold) and you can change the progress bar display, etc. isn't working properly, it seems like the menu doesn't scroll through properly using the button and it's hard to use it

Higgs2209:
Proposed fix in #73

Issue #63, opened 2026-05-27
Status screen doesn't refresh on button press and doesn't exit on double press


gnatpat:
On my device:
 - click then hold in reader opens status screen
 - clicking doesn't do anything
 - double clicking doesn't either
 - it does eventually exit, but i'm not sure what i did
 - status bar still changes, but isn't reflected in the screen

Higgs2209:
Is this the first time opening the reader menu for that book/since flashing?

Does the same behaviour occur routinely, or does it self resolve after opening reader menu again?

gnatpat:
It seems to occur on all my books, every time. Will see if I can debug if it's only me!

Higgs2209:
Proposed fix in #73 - Needs more testing

Issue #55, opened 2026-05-25
Status screen suggestions


kevinst1r:
Instead of having the in book info screen option for buttons be Menu. I suggest something like 'Book Info' or 'Book Stats' or something.  I feel 'Menu' is easy to misinterpret as Main Menu.

Also, Sidebar: Full is confusing without context.  Apparently you are supposed to change the style, but there is no setting for it.  And if it is a setting in WebUI, I feel having the Statusbar mode displayed while it is open is redundant.

Also, single click 'cycle' does nothing with the info screen open.

gamberoillecito:
I also feel that this should be a setting for the webui. If on-device settings are supposed to be the ones that you change more often I'd rather be able to cycle through font sizes rather than the style of the bar.

PR #73, opened 2026-05-28
Reader menu fixes

Higgs2209:
Issues #71 #63 

Reader menu (click-hold in a book) was unusable: clicks seemed to do nothing, the menu closed on its own, statusbar changes didn't show, and double-click-to-close froze the device.

Cause: changing the statusbar mode re-paginated the whole book on the click handler — seconds of blocked main loop. Queued clicks got misread as double/triple presses (menu self-closing); close itself froze.

Fix:

Defer repagination to the next render — in-menu clicks stay responsive.
Make repagination O(1): the current page's byte offset is layout-invariant, so keep existing offsets as back-history, truncate the table forward, and re-flow lazily. No book re-walk on close.
Force a full e-ink refresh on menu open/close so the overlay doesn't ghost.
Tested: flashed to Wireless Paper v1.2 — menu responsive, statusbar cycles with live feedback, double-click closes instantly even deep in a large book.

### DIFF
--- a/Pala_One_2_1/src/ui/reader.cpp
+++ b/Pala_One_2_1/src/ui/reader.cpp
@@ -30,17 +30,18 @@ void forceNextRenderFull() { s_forceNextFull = true; }
 // Deferred-repagination flag. Set by `markPagesDirtyForLayoutChange` when a
 // mid-session layout change (currently only Statusbar mode cycling) has
 // invalidated the in-memory page offset table. Consumed at the top of
-// `prepareForRender` so the slow re-walk happens once, on the next reader
-// draw, rather than synchronously inside the setting cycle.
+// `prepareForRender` so the rebase happens once, on the next reader draw,
+// rather than inside the setting cycle.
 //
-// Why lazy: `repaginateForLayoutChange` may walk hundreds of pages forward
-// to relocate the current byte offset under the new layout. Doing that
-// inside `Statusbar::setMode` blocks the main loop for hundreds of ms while
-// the reader menu is open — short clicks queue up in the ISR buffer, the
-// classifier groups them into Double/Triple, and the menu closes from what
-// the user thought was a single cycle. Deferring to the next render moves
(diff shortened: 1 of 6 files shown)


gnatpat:
I think this works as a solution - but I also wonder whether we need this added complexity? Or if moving this option to the Web UI settings page would be better? We would also have to key the page cache by the number of lines on a page (or the available Y space for text) - but that follows the existing patterns better and doesn't lead to a potentially long re-pagination while reading a book. Thoughts?

gamberoillecito:
> I think this works as a solution - but I also wonder whether we need this added complexity? Or if moving this option to the Web UI settings page would be better? We would also have to key the page cache by the number of lines on a page (or the available Y space for text) - but that follows the existing patterns better and doesn't lead to a potentially long re-pagination while reading a book. Thoughts?

Similarly to what I commented on #55, I think that leaving this settings to the web ui would leave more space for other information/settings on the device itself. At least for me, changing the style of the bar is something that I set and forget, I don't see it being something that deserves it's own readily accessible setting. 

I don't know how complex it would be but if someone has the need to change the style of the bar from the device itself it could be cool to add a customizable "Quick setting" to the web UI. Something like:

WebUI -> Choose "Quick setting" -> Font | Font size | Status bar style | ...

In this way the "cycle" action of the reader menu cycles through the available options for the user-selected setting.

ncont:
I agree as well to move setting to the web UI. Altho the reading statistics should stay there. When the par is minimal, it is useful to see the current page in the reader menu. 

@gnatpat , @gamberoillecito  and @Higgs2209 , do you agree to merge that fix here for v3.0 and enhance the WebUI in v3.1?

Higgs2209:
@ncont Happy to merge and leave the rest to 3.1.

I am going to spend some time bug fixing tomorrow and if I have time before 3.1 I'll try and get this one live.

Agree to merge this.

Issue #47, opened 2026-05-25
Delete button size too small


kevinst1r:
In screensaver editor the delete button is tiny and should be the same size as download.

PR #75, opened 2026-05-29
Fix screensaver delete button size to match download

Higgs2209:
Simple fix to fix the delete button size issue #47


ncont:
Tested: Worked

image

PR #78, opened 2026-05-30
Enable cppcheck

ncont:
# chore: enable cppcheck static analysis CI + fix all reported findings

  ---
  Summary
  
  This branch introduces automated static analysis via cppcheck and resolves every finding it surfaces on the
  codebase.

  CI workflow (static-analysis.yml): runs pio check on every push and pull request targeting any branch, failing on
  any finding of severity medium or higher. One PlatformIO environment is enough since all build targets share the
  same source tree.
  
  Code fixes span the following categories:

  knownConditionTrueFalse — dead conditions removed

  - input.cpp: Simplified press/release checks — after the edgePressed == stablePressed_ continue-guard, the two
  values are always opposite, making the redundant half of each condition trivially true/false.
  - input.cpp: Replaced else if (clickCount_ >= 4) with else — the preceding == 1/2/3 branches exhaust the only
  alternatives.
  - font.cpp: Removed if (w < 50) w = 50 — with SCREEN_W=250 and MARGIN_X=6, w is always 238.
  - bookmark_list_screen.cpp, web/bookmarks.cpp (×2): Removed if (targetPage/storedPage < 0) guards — pages[] is
  uint16_t, so the cast to int is always ≥ 0.
  - battery.cpp: Inside if (pctRaw < pctShown), the inner (pctShown - pctRaw) >= 1 is always true for integers;
  collapsed to a direct decrement.

  constVariablePointer / constParameterPointer — const-correctness
  
  - app_loader.cpp: relocs pointer only reads the relocation table, not writes — declared const uint32_t*.
  - paginator.cpp: trimTrailing lambda only reads buf — parameter changed to const char*.

  uninitvar — uninitialized array

  - paginator.cpp: token[kTokenMax] zero-initialized at declaration (= {}). cppcheck couldn't prove the array was
  always filled before the hardBreakToken lambda reads it across captured references.

  unreadVariable — dead assignments removed

  - bookmark_label.cpp: Post-loop if (inWord) words++ — words is never read after the loop exits.
  - statistics_screen.cpp: Last y += lineH after the final u8g2.print — y is unused thereafter (the bitmap section
  below uses its own yTop).
  - upload_screen.cpp (×2): Trailing y += at the end of each branch of the station/AP if-else — y is not read before
  display.update().
  - web/settings.cpp: hasSleepImg was computed but never referenced anywhere in the output.

  shadowFunction — local variable naming conflicts

  - bookmark_list_screen.cpp, web/bookmarks.cpp (×3), web/files.cpp: Local String bookPath variables renamed to
  filePath to avoid shadowing the global bookPath(int idx) function from library.h.

  unusedFunction — suppress comments for cross-TU functions

  Added // cppcheck-suppress unusedFunction to functions that are called from other translation units but invisible
  to single-TU analysis:
  - header_title.cpp: loadSettings, current, set
  - reader.cpp: markPagesDirtyForLayoutChange
  - widgets.cpp: forceNextMenuFrameFull, prepareMenuFrame
  - text_util.cpp: utf8CharAt
  
  Test plan

  - [x] CI passes on this branch (pio check -e wireless-paper-v1_2-en --fail-on-defect medium exits 0)
  - [x] No behavioral changes — all fixes are either dead-code removal, const qualification, naming, or suppress
  annotations


gnatpat:
Why do we need all the `// cppcheck-suppress unusedFunction`? Is there a way to turn it off globally?

ncont:
> Why do we need all the `// cppcheck-suppress unusedFunction`? Is there a way to turn it off globally?

Cpp check can not look past its scope and wants do check globally. Which is apparently not always possible. There seems so be multiple threads trying to fix it (https://trac.cppcheck.net/ticket/9682, https://trac.cppcheck.net/ticket/10727) 

Recommendation is to surpress it. Question is now: globally or locally. 

I thought locally is better to actively decide for it and be able to catch the real dead functions. Especially when the air refractors code. My agents try to forget things from time to time.

But: if you don't like it, it can turn it off. I want it to be used and not annoy people:D

ncont:
Ok. the unusedFunction annoyed me already. Removed it.

Higgs2209 (approved):

PR #84, opened 2026-05-31
File Round Trip

ncont:
### Summary

  Completes a full upload/download round-trip for every file type managed by the device, and promotes the Screensaver
  editor to a top-level navigation entry.

  ---

  ### Books

  - Added `GET /download-book?id=N` route that streams the stored `.txt` file as an attachment
  - Download button added to each book entry in `/files`, stacked above the existing Delete button

  ### Apps

  - Added `GET /download-app?name=foo.bin` route that streams the installed `.bin` as an attachment
  - Download button added to each app entry in `/files`, same layout as books
  - Both routes validate the filename (rejects path traversal, enforces `.bin` extension) before opening the file

  ### Screensavers

  - Added raw `.bin` upload to every rotation slot — an upload icon button (arrow-up, mirroring the existing
  arrow-down download icon) appears on each slot card, including empty ones
  - Clicking the icon opens a native file picker filtered to `.bin`; selecting a file auto-submits to the existing
  `/screensavers/upload?slot=N` endpoint, which already validates the 3904-byte size
  - Same upload button added to the Single Screensaver card, posting to `/screensavers/upload?single=1` (saves as
  `/sleep.bin`)
  - Download + Delete icons on a slot or single image are only shown when the slot is populated; Upload is always
  visible

  ### Navigation
  
  - **Screensaver** added as a top-level nav entry on the home page, `/files`, and `/settings`
  - Screensaver card removed from the Settings page (it was just a link button to `/screensavers`)
  - Screensaver page nav updated from `← Home` to `Home | Settings` — it is now a peer-level page, not a sub-page of
  Settings
  - Dead string constants (`D_WEB_SCREENSAVER_CARD_DESC`, `D_WEB_SCREENSAVER_EDITOR_LINK`,
  `D_WEB_SCREENSAVER_EDITOR_HINT`) removed from both language files

  ---
  
  ### Round-trip status after this PR

  | Type | Upload | Download |
  |---|---|---|
  | Books | ✅ `.txt` via home page form | ✅ `.txt` via Download button in `/files` |
  | Apps | ✅ `.bin` via home page form | ✅ `.bin` via Download button in `/files` |
  | Screensavers | ✅ image via editor · ✅ `.bin` via slot upload icon | ✅ `.bin` per slot / single |


Higgs2209 (approved):
Happy for this to be merged based on code review.

Will test on device this afternoon and is all is good will merge.

PR #54, opened 2026-05-25
local web flasher for dev purposes

tomascorreag:
## Summary

Adds a local web flasher tool that lets developers build and flash any branch or fork directly from the browser — no CI or gh-pages deployment needed.

- **`scripts/local_flash.py`** — Run `python scripts/local_flash.py` to open the installer page locally. A branch selector in the browser lets you pick any local/remote branch (or paste a GitHub URL to a public fork), choose a board variant, and click Build. The firmware is built in an isolated `git worktree`, then served on localhost for flashing via Web Serial.
- **`scripts/assemble_site.py`** — Added `--lenient` flag: skips missing firmware variants instead of aborting. Variant HTML blocks are now generated dynamically from metadata instead of being hardcoded in the template. Also fixes a latent `UnicodeDecodeError` on Windows (UTF-8 template read with default cp1252 encoding).
- **`install/index.html`** — Hardcoded variant blocks replaced with `{{VARIANTS}}` placeholder (output is identical when all 4 variants are built). Added branch selector UI and local-dev walkthrough, both hidden on hosted channels via `body:not(.channel-local)` CSS. Added client-side manifest check that disables install buttons when no firmware is available. Hosted (stable/dev) page content is unchanged.

## CI impact

None. The workflow, manifests, and landing page are untouched. `assemble_site.py` in strict mode (CI default, no `--lenient`) produces identical output — the dynamic variant generation matches the previous hardcoded HTML. The `index.html` template additions (branch selector, local walkthrough) are hidden via CSS on non-local channels.

## Test plan

- [x] `python scripts/assemble_site.py --version test` (strict mode, all 4 envs built) — should succeed, output identical to current behavior
- [x] `python scripts/assemble_site.py --lenient --version test` (partial build) — should succeed, only include available variants
- [x] `python scripts/local_flash.py` — opens browser at localhost:8080, shows branch selector and "No build available" if no prior build
- [x] Select a branch, pick a board, click Build — builds in worktree, page reloads with install button enabled
- [x] Paste a GitHub URL (e.g. `https://github.com/PaulLagier/pala-one-firmware/tree/dev`) — fetches and builds from the remote fork
- [ ] Hosted page (`--channel dev`) — branch selector hidden, upstream walkthrough and "Not sure which version" note visible, no visual changes

PR #85, opened 2026-06-01
Serve PR builds on web-flasher through CI

Roanmh:
## Problem

While installing the dev build is trivial with the web flasher, testing a PR
requires checking out and building from source.

## Proposal

Each PR creates a page on the web flasher when the builds finish. A comment
directs testers to the flash page.

This is solving a slightly different problem than #54, but the PR's may conflict. If we agree this is a useful idea. I am happy to work with @tomascorreag propose a reconsiliation of the ideas.

## How it works

There is a subtle separation between what code runs with what privledges to keep the repo safe. Here's my understanding, after learning this from claude:
1. CI code runs in a controlled environment, so the resulting build can be trusted to match the source code of the diff. Testers should either trust the poster or trust their understanding of the changes before flashing.
2. The actions are separated and communicate through artifacts to limit permissions of untrusted code. Detailed below.

**`pr-build.yml`** — triggered by `pull_request`. Builds all variants,
assembles the installer site, and uploads it as a workflow artifact. Uses the
read-only `pull_request` token so fork code is never executed with write
access.

**`pr-deploy.yml`** — triggered by `workflow_run` once the build completes.
Runs in the base-repo context (write access), downloads the artifact, deploys
to `gh-pages/pr/<N>/` alongside the existing `/dev/` and `/stable/` channels,
then posts a comment linking to the flash page. No fork code is executed here.

**`pr-cleanup.yml`** — triggered by `pull_request_target` on PR close.
Deletes `pr/<N>/` from `gh-pages`. No code checkout.

The `workflow_run` split is what makes fork PRs safe: untrusted code only runs
under the read-only `pull_request` token; the privileged deploy step never
touches the fork's code.

**`install/index.html`** gets a `pr` channel CSS class (orange badge, same
visual as `dev`) and a warning banner that is only shown on PR builds. No
changes to `assemble_site.py` — `--channel pr` is already valid.

The installer URL will be
`https://paullagier.github.io/pala-one-firmware/pr/<N>/`.

## Not included in this PR

- Pending "building…" comment while the build runs
- Failure comment if the build errors
- Parallel matrix builds (currently sequential, ~6 min total)

These are straightforward follow-ons once the core is merged.

## Testing

Live at https://github.com/Roanmh/pala-one-firmware/pull/2

TODO: test plan once we agree on value and architecture


gamberoillecito:
I believe this would be quite useful to make beta testing of new features easier and hopefully speed up development.

ncont:
Please reopen as soon as you have updates. I will close the stale PR for now to clean up the PR list.

PR #91, opened 2026-06-04
Time Statistics Addition

### LINKED ISSUE #79: Add time-based statistics
I think it would be nice to add all or some of the following statistics to the current ones:
- Today's reading time
- Week's reading time
- Monthly reading time
- Yearly reading time
- Reading time since last charge
- Average reading time per day|week|...

Given the large number of additional information the current view should be changed to allow cycling through different pages.

Higgs2209:
## Reading-time statistics

in response to #79 

Adds reading-time tracking to the statistics screen.

- Today / week / month / year reading time + average per day, on a new second stats page.
- Time accrues from the gap between page turns; gaps over 5 min are treated as idle and dropped. State lives in RTC RAM and flushes to NVS on the existing safety-net + sleep path, so frequent page turns add no flash wear. New `cfg_rtime` blob; buckets use the streak day-index model.

**UI / controls**
- Statistics screen is now two pages: streak/lifetime (1) and reading time (2).
- One click cycles pages; double-click exits to the library.

**Tests**
- Pure bucket logic in `pure/reading_time_log.{h,cpp}`, host-tested (`test_reading_time_log.cpp`).
- Strings mirrored across `en.h` + `es_la.h`; full firmware build (`wireless-paper-v1_2-en`) passes.

### DIFF
--- a/Pala_One_2_1/src/lang/en.h
+++ b/Pala_One_2_1/src/lang/en.h
@@ -47,6 +47,16 @@
 #define D_STATS_LIFETIME_PAGES_FMT     "Pages turned: %llu"
 #define D_STATS_LIFETIME_PRESSES_FMT   "Button presses: %llu"
 
+// Reading-time page (page 2 of the statistics screen). %s is a duration
+// formatted by the screen (e.g. "1h 23m" or "45m").
+#define D_STATS_TIME_HEADING           "Reading time"
+#define D_STATS_TIME_TODAY_FMT         "Today: %s"
+#define D_STATS_TIME_WEEK_FMT          "Week:  %s"
+#define D_STATS_TIME_MONTH_FMT         "Month: %s"
+#define D_STATS_TIME_YEAR_FMT          "Year:  %s"
+#define D_STATS_TIME_AVG_FMT           "Avg/day: %s"
+#define D_STATS_BACK_HINT              "Click to go back"
+
 // ----------------------------------------------------------------------------
 //  List screen (src/ui/screens/list_screen.cpp)
 // ----------------------------------------------------------------------------
--- a/Pala_One_2_1/src/lang/en.h
+++ b/Pala_One_2_1/src/lang/en.h
@@ -55,7 +55,6 @@
 #define D_STATS_TIME_MONTH_FMT         "Month: %s"
 #define D_STATS_TIME_YEAR_FMT          "Year:  %s"
 #define D_STATS_TIME_AVG_FMT           "Avg/day: %s"
-#define D_STATS_BACK_HINT              "Click to go back"
 
 // ----------------------------------------------------------------------------
 //  List screen (src/ui/screens/list_screen.cpp)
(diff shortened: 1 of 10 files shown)


Pala_One_2_1/src/pure/reading_time_log.h
```diff
@@ -0,0 +1,66 @@
...
+
+static const uint32_t RTIME_SCHEMA         = 1;
+static const uint32_t RTIME_DAYS_PER_WEEK   = 7;
+static const uint32_t RTIME_DAYS_PER_MONTH  = 30;
+static const uint32_t RTIME_DAYS_PER_YEAR   = 365;
+
+// Max seconds credited to a single page-turn gap. Longer gaps are treated as
+// "put the book down" (idle / overnight / first turn after deep-sleep wake)
+// and dropped entirely rather than clamped, so reading time never counts idle.
+static const uint32_t RTIME_GAP_CAP_SECS = 300;  // 5 min/page ceiling
```
ncont, line 29: If i get it correctly, than this conflicts with the sleep time. The default sleep time is 120sec. 

I would maybe cap the time when we enter sleep and flush it. 

The whole idea of the reading time is, that we expect the user reads to the very last sec until the screensaver comes. but that all we can do.

A short notice to the user in the Tooltip of the "no-screensaver" feature would be good, that the reading time is not counted.

Pala_One_2_1/src/lang/en.h
```diff
@@ -47,6 +47,15 @@
 #define D_STATS_LIFETIME_PAGES_FMT     "Pages turned: %llu"
 #define D_STATS_LIFETIME_PRESSES_FMT   "Button presses: %llu"
 
+// Reading-time page (page 2 of the statistics screen). %s is a duration
+// formatted by the screen (e.g. "1h 23m" or "45m").
+#define D_STATS_TIME_HEADING           "Reading time"
```
ncont, line 52: I think it is pretty hard to detect "reading".  Especially if a user (like me), use the "no-screensaver" feature. 

We can actually only measure the time the device is turned on. 

"Acive for:" or "Used for: "

Something like that. But it is just an idea.

ncont (commented):
After i read this, i think that it is pretty hard to detect "reading". 

What do you think by naming it "Active for:

Pala_One_2_1/src/pure/reading_time_log.h
Higgs2209: Yes this is a good point. Thanks for your review. I will implement

Pala_One_2_1/src/lang/en.h
gamberoillecito: Can't you do `reading_time ≈ time_on - time_in_menues - autosleep_time`? I use "no screensaver" too but the time I spent with the screen locked is the time needed for reading a single page («1 minute) for every reading session. Please let me know if I'm missing some other scenario.

Also displaying "active/used for" would led me to believe that the displayed time accounts also for the time spent in the menus/connecting to the webUI and so on.

gamberoillecito:
Hi, what happened to this PR?

Higgs2209:
> Hi, what happened to this PR?

Arghh.... I was cleaning up my branches and much of deleted this branch accidentally....

I will fix after work

Issue #95, opened 2026-06-05
Simplify battery charge indicator


gamberoillecito:
After discord discussion https://discord.com/channels/1502006584639881256/1504853975143551226/1512490671171436837 I propose changing the battery indicator to:

<img width="1276" height="1552" alt="Image" src="https://github.com/user-attachments/assets/94ae0494-4932-4a22-a8c5-9fcd0e478cf7" />

This issue should be resolved together with or after #67

gamberoillecito:
Solved by #97

Issue #67, opened 2026-05-27
Battery indicates 100% when charging


ncont:
A user is unsettled because the battery indication shows 100% when charging and after unplugging the charge drops to the actual value (for example 60%) 

https://discord.com/channels/1502006584639881256/1504853975143551226/1509249510843486258

Resoning: 

We measure the battery voltage by the voltage across the battery. This is just a poor mans solution. The assumption is, that the voltage drops somewhat linear. Ofc that is not true in reality. But good enough to let the user know to charge the device.

To get the actual battery "State of Charge" you would need to add a coulomb counter. With that you can measure what went out and what do you have to put back in to reach 100%. But the Heltec does not have such a counter.

Now, what would have been great is to know if the USB C is in and then we could display a charging icon. And the USB C gives out the VUSB. But it is not routed to a GPIO.

Last Idea is to check the foltage if it is capped at 100% or even higher which should indicate charging as the voltage needs to be higher than the Battery voltage. This is a poor mans solution as well tho.

When charging, the Battery Icon should not display the percentage but a Battery with a flash in it.

gamberoillecito:
I'd like to try working on this if no one is already on it, no promises though ;)

gamberoillecito:
Small update on my mostly failed attempts:

I could not find any electrical signal to use to measure the battery charge more accurately that what is being done right now, so I set the problem aside and tried working on detecting if the battery is charging to display a charging icon instead of the misleading 100% charge.

I tried three different approaches for detecting if the charger is connected:

1. `WifiProvisioning::isActive()`: gave mixed results, showing that the board was connected to USB only when connected to the PC USB port with the PC turned OFF (I did only few tests so it may have been a fluke)
2. `usb_serial_jtag_is_connected()`: seemed promising but never actually worked. https://docs.espressif.com/projects/esp-iot-solution/en/latest/usb/usb_overview/usb_serial_jtag.html
3. Check if voltage reading gives a value higher than 100%. This method seemed to work reliably in my limited tests (i.e. always reported correctly when the board was plugged in).

No matter the method, I faced an additional issue: how to update the charging icon as soon as the charger is connected/disconnected.

Currently the icon is updated every `BAT_CACHE_MS` (3 minutes) and I don't know how to trigger an update as soon as the charger is connected. I would prefer not to decrease because I fear it may have negative impact on the battery life. 

Any suggestion is welcome as I don't have many other ideas to test

ncont:
Good work!

[...]

PR #97, opened 2026-06-06
Battery UI improvements

gamberoillecito:
This PR aims at solving issues #67 and #95. The following high level modifications have been made:

1. Charging detection based on the threshold voltage `BAT_CHARGING_VOLTAGE`
2. Periodically check if the battery is charging every `BAT_CHARGING_CHECK_MS`, and update `s_battery` in `battery.cpp`. If a charging state change is detected AND the battery indicator is currently shown* update the battery indicator accordingly. Otherwise fall back to a less frequent update as before.
3. Simplify the battery charge indicator as proposed in #95.

*: Currently this is based on the hardcoded logic in `batteryIndicatorVisible()` (Pala_One_2_1.ino). 

These changes may need several days to properly test due to the long battery life of the device. I tried to simulate different battery levels and I couldn't find significant issues but real use may show different bugs.

### DIFF
--- a/Pala_One_2_1/src/hal/battery.cpp
+++ b/Pala_One_2_1/src/hal/battery.cpp
@@ -197,6 +197,70 @@ bool batteryChargingChanged()
   return changed;
 }
 
+// Draws the outline of a battery with top-left corner at (x, y), width `w` and height `h`
+void drawBatteryOutline(int x, int y, int w, int h)
+{
+  gfx.drawRect(x, y, w, h, 1);             // main body
+  gfx.fillRect(x + w, y + 2, 2, h - 4, 1); // positive terminal on the right
+}
+
+// Draws a solid bar indicating `perc` percentage of the battery charge. 0<= perc <= 100.
+// The coordinates refer to the top-left corner of the battery
+// `w` and `h` are the width and height of the battery
+void drawBatteryCharge(int battX, int battY, int battW, int battH, int perc)
+{
+  if (perc == 0)
+  {
+    return;
+  }
+
+  int fillW = perc * (battW - 2) / 100;
+  gfx.fillRect(battX + 1, battY + 1, fillW, battH - 2, 1);
+}
+
+// Draws an exclamation mark leaving `spacing` pixels between it and the LEFT of the battery.
+void drawExclamation(int battX, int battY, int battH, int spacing)
+{
+  int exMarkX1 = battX - spacing - 2;
+  int exMarkX2 = battX - spacing - 1;
+  gfx.fillRect(exMarkX1, battY, 2, battH - 4, 1);     // "pipe" part of the exclamation mark
+  gfx.fillRect(exMarkX1, battY + battH - 2, 2, 2, 1); // "dot"
+}
+
+void drawChargingFill(int battX, int battY, int battH, int battW)
+{
+  gfx.fillTriangle(battX, battY, battX, battY + battH - 1, battX + battH, battY, 1);                                             // left triangle
+  gfx.fillTriangle(battX + battW - 1, battY, battX + battW - 1, battY + battH - 1, battX + battW - battH, battY + battH - 1, 1); // right triangle
+}
+
+// Draws a bolt symbol leaving `spacing` pixels between it and the LEFT of the battery.
+//    a
+//   /
+// b -- c
+//    /
+//   d
+
+void drawBolt(int battX, int battY, int battH, int spacing)
+{
+  int a_x = battX - spacing;
+  int a_y = battY;
+  int b_x = battX - spacing - battH / 2;
+  int b_y = battY + battH / 2;
+  int c_x = a_x - 1;
+  int c_y = b_y;
+  int d_x = b_x - 1;
+  int d_y = battY + battH - 1;
+
+  int thickness = 2;
+
+  for (int i = 0; i < thickness; i++)
+  {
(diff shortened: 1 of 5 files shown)


Higgs2209 (approved):
Code makes sense. Will flash and time will tell how well it works in practice

gamberoillecito:
I performed some tests exploiting #98 to drain the battery faster than normally, going from full to very low. The update seems to work as expected except for the following:

I noticed something unexpected, which could also be unavoidable due to the charge reading method we are using. If the battery is low (i.e. < 25%), plugging the charger for a couple of minutes brings the battery to full charge (according to the indicator implemented in this PR).

gamberoillecito:
I added a small debug info at the top of the screen showing the current battery voltage reading to make it easier to see what is going on. This should be removed before merging.

gamberoillecito:
New day new bug: plugging the charger when the battery is nearly empty (~3.4V) makes the voltage ramp up slowly, it takes a few minutes to reach the charging threshold so the charging icons does not appear for a while. Let me know if you can reproduce this

Pala_One_2_1/src/hal/battery.cpp
```diff
@@ -62,112 +75,284 @@ static uint32_t readAdcMilliVoltsStable() {
...
+
+  int fillW = perc * (battW - 2) / 100;
+  gfx.fillRect(battX + 1, battY + 1, fillW, battH - 2, 1);
+}
+
+// Draws an exclamation mark leaving `spacing` pixels between it and the LEFT of the battery.
+void drawExclamation(int battX, int battY, int battH, int spacing)
+{
+  int exMarkX1 = battX - spacing - 2;
+  int exMarkX2 = battX - spacing - 1;
```
ncont: [low:style] Variable 'exMarkX2' is assigned a value that is never used. [unreadVariable]

ncont:
I would like to draw the battery full when charged. Same behavior as with the exclamation mark. There you also just add the mark. 

I see the instability of the battery reading with mine as well. I used a cheap low power charger and the clicking through the menu drained the battery faster then it charged. Which resulted that it was not showing power but fully charged. 

I would add two things: 
- A text in the Readme explaining the limitations. I don't see any better way rn. 
- Since a user asked to leave the battery percentage, and i actually like the voltage reading, we could add this information (percentage and voltage) in the "Device" Screen (i think it is called about screen in the code).

After you added it there and removed the debug text, i would merge it and see how it will behave over a longer period.

gamberoillecito:
@ncont thanks for the feedback, could please clarify this:

> I would like to draw the battery full when charged. Same behavior as with the exclamation mark. There you also just add the mark.

I don't understand what you would like me to edit/add.

Regarding the changes in the "Device" screen, my only concern is that they wouldn't update "live" and I don't know if it makes sense refreshing the screen every `x` seconds. Do you think it would be ok to update the values only when the "Device" screen is opened?

Side note: why have all checks failed?

ncont:
> @ncont thanks for the feedback, could please clarify this:
> 
> > I would like to draw the battery full when charged. Same behavior as with the exclamation mark. There you also just add the mark.
> 
> I don't understand what you would like me to edit/add.

When the charge icon is drawn, there is a thunder next to the battery. The battery is not fully back but parts plans and some empty. Don't know how to describe it :D. I would like a solid black battery with a thunder next to it. 


> Regarding the changes in the "Device" screen, my only concern is that they wouldn't update "live" and I don't know if it makes sense refreshing the screen every `x` seconds. Do you think it would be ok to update the values only when the "Device" screen is opened?

Ok for now. I would like to update that screen with more infos like disk space. And then I will change the behavior a bit to have a update function. But that would be to much for this PR. 

> Side note: why have all checks failed?

Because 'pio check' fail due to the unused function. Fix https://github.com/PaulLagier/pala-one-firmware/actions/runs/27075439237/job/79981458476#step:6:45 and it will pass.

gamberoillecito:
> I would like a solid black battery with a thunder next to it.

Ok I understand, the reason why I didn't want to put the completely filled battery because I think you can get the wrong impression that the battery is fully charged even though you just plugged the charger in

ncont:
> > I would like a solid black battery with a thunder next to it.
> 
> Ok I understand, the reason why I didn't want to put the completely filled battery because I think you can get the wrong impression that the battery is fully charged even though you just plugged the charger in

Ah yes. Now I can remember that you mentioned somewhere. Sorry never mind then :)

gamberoillecito:
@ncont I updated the readme, fixed the error with pio check and removed the debug info. I haven't updated the "Device" screen though for the following reasons:
1. The space on the current screen is quite limited at the moment and it would need a bit of work to support multiple screens.
2. The current information are hard-coded (see `D_ABOUT_GESTURE_NEXT` and similar in `config.h` and `about_screen.cpp`), meaning a bit more of work.

If you agree we can open a new issue to improve on this screen and leave this PR as is

ncont:
> @ncont I updated the readme, fixed the error with pio check and removed the debug info. I haven't updated the "Device" screen though for the following reasons:
> 1. The space on the current screen is quite limited at the moment and it would need a bit of work to support multiple screens.
> 2. The current information are hard-coded (see `D_ABOUT_GESTURE_NEXT` and similar in `config.h` and `about_screen.cpp`), meaning a bit more of work.
> 
> If you agree we can open a new issue to improve on this screen and leave this PR as is

Yes, good decision. I hoped there is some space left and a simple modification.

PR #102, opened 2026-06-07
Add support for flipping screen orientation

gamberoillecito:
Added support for flipping screen orientation both in webUI and as an action which can be mapped to one of the device's clicks.

Motivation:
1. Support for different case designs
2. I would like to be able to flip the device while reading so that I can hold it in a different way to avoid fatigue

### DIFF
--- a/Pala_One_2_1/Pala_One_2_1.ino
+++ b/Pala_One_2_1/Pala_One_2_1.ino
@@ -111,6 +111,7 @@
 #include "src/ui/text.h"
 #include "src/ui/toast.h"
 #include "src/web/web.h"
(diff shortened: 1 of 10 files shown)


ben-igel:
has this 90 or 180 degrees rotation?

gamberoillecito:
> has this 90 or 180 degrees rotation?

180 only, I think that vertical aspect ratio would be mostly unusable and require various changes to the rest of the code. With this PR I basically only changed 2 lines of code and adjusted the settings and necessary boilerplate

ben-igel:
oh that's sad but understandable

Issue #80, opened 2026-05-31
Uniform click patterns


gamberoillecito:
Currently, when reading a book, 1 click corresponds to forward and 2 clicks to backward but this behaviour is not consistent across the device.
In the main menu 2 clicks correspond to select.
In some of the default "apps", like "upload" 1 click goes back to the home screen.

Proposed modes to be selected from webUI;

1. Legacy mode: keeps the current behavior
2. New mode (name to be improved):
   - 1 click: NEXT. This should iterate through menu items (in the main menu) or pages (in books or multi-page screens like the current book menu or for example what proposed in #79)
   - 2 clicks: PREVIOUS. The opposite of NEXT
   - 3 clicks: HOME. As it is currently
   - long click: ACTION. Depending on the current screen. In the main menu it should select the currently focused item (same as the 2 clicks in the current implementation). In the other circumstances it can be adapted to the different needs.
   - If a given option is not available, the corresponding click pattern should do nothing. E.g. in the "upload" app which only has one page, 2 clicks should not go to the main menu.

ncont:
I actually agree to that double click should select the previous item in menu. With more books and menu entries, it is hard to only go forward. 

But: slecting an item should be three clicks. Otherwise is takes ages to confirm. 

And then long press could be "return to menu"

Anyway: Lets discuss this when we plan the next release!

ben-igel:
yes! going up in a list with double-click would be nice and logical. 
and if long-press is less than a second it would be perfect. triple-click isn't fun to use. maybe tap+hold for menu? and tap-tap-hold could be assigned too (everything could be adjusted in the webinterface of course)

gamberoillecito:
I think the best approach would be to separate the specific pattern from the action similarly to what remapping controllers in games does. I still have to check the code though and I'm no expert so this may be either very complex or already implemented 😅.

With this approach the actions could be hard-coded (e.g. NEXT, PREVIOUS, HOME, SELECT/ENTER, CUSTOM, ...), as well as the click patters (single, double, triple, ...). Then in the config (or webui) the user can assign each pattern to each action. The programmer instead doesn't need to worry about clicks but rely only on the action.

Please let me know if this makes sense or if you think this would be feasible to implement

ben-igel:
can't wait for going up in lists!

gamberoillecito:
I'm working on this

gamberoillecito:
[...]

Issue #112, opened 2026-06-23
Move progress bar settings to web ui


gamberoillecito:
Following https://github.com/PaulLagier/pala-one-firmware/pull/73#issuecomment-4581144901 and #110 I think it would be better to move the settings for the progress bar to the web ui.

gamberoillecito:
In case #123 is merged, it may nonetheless make sense to move this to the on-device layout settings and clean up the "book menu"

PR #114, opened 2026-06-28
Uniform actions

gamberoillecito:
This PR addresses #80 by allowing the user to remap any click pattern (i.e. gesture) to any action, giving the option to enable a legacy mode via the webUI to keep the old behaviour.

## Philosopy

I tried to strike the correct balance between making the experience usable and exploiting the new options given by the new remapping (i.e. going up in the main menu), while limiting new features/behaviours where not fundamental at the moment. Such features are mostly quality of life improvements (e.g. looping from the first to the last main menu entry). This was done with the goal of having a way to test the functionality and leaving space for a future PR that refines everything in a way that makes sense for the community.

## Web UI

The web ui was improved to accomodate the new features, with the addition of server side validation to make sure that the user doesn't remain locked out of the system (e.g. if they forget to remap the "HOME" action). This is done via a js `alert` for simplicity, a more polished solution may be added in the future.

## Legacy mode

Currently, legacy mode is enabled by default to avoid unsettling users who update without knowing about the new changes. It would be nice to keep it this way for those who update but turn legacy mode off by defaults for new installs. I don't know how to achieve this though.

## Planned* implementation

| Action   | Main menu effect   | Submenu effect (e.g. "connect") | Reading            | Locked | Default gesture binding |
|----------|--------------------|---------------------------------|--------------------|--------|-------------------------|
| NEXT     | Next item          | Next page (if multi-page)       | Next page          | /      | Short click             |
| PREV     | Previous item      | Previous page (if multi-page)   | Previous page      | /      | Double click            |
| OK       | Select item        | Iteract/Cycle options           | Book menu          | /      | Long click              |
| LOCK     | Lock               | Lock                            | Lock               | Unlock | Extra long click        |
| HOME     | Go to first item   | Go to previous menu             | Go to main menu    | /      | Triple click            |
| BOOKMARK | /                  | /                               | Bookmark page      | /      | Click + Hold            |
| ROTATE   | Flip screen 180deg | ?                               | Flip screen 180deg | /      | /                       |

*: due to the reasons explained above and detailed in my following comment, not all behaviours have been implemented


gamberoillecito:
I'm also taking the opportunity to note things that should be changed. Let me know if they should be addressed here or later in a separate issue:

- [ ] Reader Menu: cycling between status bar styles hasn't been updated (but there has been a discussion about moving this option to the web UI: #112)
- [x] Statistics screen: a single click goes back to the main menu despite legacy mode being off
- [x] Device screen: a single click goes back to the main menu despite legacy mode being off
- [x] Firmware update screen: ACTION_PREV doesn't work
- [x] Main menu (i.e. library screen): when on the first item, ACTION_PREV doesn't loop to the last item. 
- [ ] ACTION_MENU should be renamed to ACTION_OK_MENU or similar to distinguish it from ACTION_HOME. Same goes for its label displayed to the user in the webUI and in the "Device screen".
- [ ] Similarly to the previous list item, ACTION_HOME may be renamed to ACTION_BACK or similar, together with its label. The reason for this is that in cases like when browising bookmarks, the current ACTION_HOME is used to go "up" one menu instead of straight to the main menu
- [ ] The device can currently only be unlocked by long press I believe. I haven't changed the behaviour as I don't know if this has battery life implications
- [x] In the main menu (library screen), the only recognized actions are _MENU, _NEXT, _PREV. It may be worth it to respond to also _ROTATE and _LOCK
- [x] ACTION_HOME in the main menu doesn't go to the first menu item

Higgs2209:
> I'm also taking the opportunity to note things that should be changed. Let me know if they should be addressed here or later in a separate issue:
> 
>     * [ ]  Reader Menu: cycling between status bar styles hasn't been updated (but there has been a discussion about moving this option to the web UI: [Move progress bar settings to web ui #112](https://github.com/PaulLagier/pala-one-firmware/issues/112))
> 
>     * [x]  Statistics screen: a single click goes back to the main menu despite legacy mode being off
> 
>     * [x]  Device screen: a single click goes back to the main menu despite legacy mode being off
> 
>     * [x]  Firmware update screen: ACTION_PREV doesn't work
> 
>     * [x]  Main menu (i.e. library screen): when on the first item, ACTION_PREV doesn't loop to the last item.
> 
>     * [ ]  ACTION_MENU should be renamed to ACTION_OK_MENU or similar to distinguish it from ACTION_HOME. Same goes for its label displayed to the user in the webUI and in the "Device screen".
> 
>     * [ ]  Similarly to the previous list item, ACTION_HOME may be renamed to ACTION_BACK or similar, together with its label. The reason for this is that in cases like when browising bookmarks, the current ACTION_HOME is used to go "up" one menu instead of straight to the main menu
> 
>     * [ ]  The device can currently only be unlocked by long press I believe. I haven't changed the behaviour as I don't know if this has battery life implications
> 
>     * [x]  In the main menu (library screen), the only recognized actions are _MENU, _NEXT, _PREV. It may be worth it to respond to also _ROTATE and _LOCK
> 
>     * [x]  ACTION_HOME in the main menu doesn't go to the first menu item

I am not sure if these have already been address, so I aplogise, but yes, these should all be seperate issues/PRs.

Higgs2209 (approved):

Higgs2209:
@gamberoillecito - I am happy from a purely code review perspective. I will give @ncont a day or to to see if they want to review and for me to test on device, then I will merge.

Thanks for your work on this.

gamberoillecito:
> > I'm also taking the opportunity to note things that should be changed. Let me know if they should be addressed here or later in a separate issue:
> > ```
> > * [ ]  Reader Menu: cycling between status bar styles hasn't been updated (but there has been a discussion about moving this option to the web UI: [Move progress bar settings to web ui #112](https://github.com/PaulLagier/pala-one-firmware/issues/112))
> > 
> > * [x]  Statistics screen: a single click goes back to the main menu despite legacy mode being off
> > 
> > * [x]  Device screen: a single click goes back to the main menu despite legacy mode being off
> > 
> > * [x]  Firmware update screen: ACTION_PREV doesn't work
> > 
> > * [x]  Main menu (i.e. library screen): when on the first item, ACTION_PREV doesn't loop to the last item.
> > 
> > * [ ]  ACTION_MENU should be renamed to ACTION_OK_MENU or similar to distinguish it from ACTION_HOME. Same goes for its label displayed to the user in the webUI and in the "Device screen".
> > 
> > * [ ]  Similarly to the previous list item, ACTION_HOME may be renamed to ACTION_BACK or similar, together with its label. The reason for this is that in cases like when browising bookmarks, the current ACTION_HOME is used to go "up" one menu instead of straight to the main menu
> > 
> > * [ ]  The device can currently only be unlocked by long press I believe. I haven't changed the behaviour as I don't know if this has battery life implications
> > 
> > * [x]  In the main menu (library screen), the only recognized actions are _MENU, _NEXT, _PREV. It may be worth it to respond to also _ROTATE and _LOCK
> > 
> > * [x]  ACTION_HOME in the main menu doesn't go to the first menu item
> > ```
> 
> I am not sure if these have already been address, so I aplogise, but yes, these should all be seperate issues/PRs.

Hi, I addressed the more related ones (the ones marked in the list) and left the others for later. When using the device I felt those issues needed to be addressed to make the experience usable (e.g. not knowing if double click would go to the previous item or select was extremely frustrating)

ncont:
> @gamberoillecito - I am happy from a purely code review perspective. I will give @ncont a day or to to see if they want to review and for me to test on device, then I will merge.
> 
> Thanks for your work on this.

Looks good to me as well. @gamberoillecito , after your rebase, we can merge it.

gamberoillecito:
@ncont it should be ready to merge, I don't have much experience with rebasing so let me know in case I did something wrong

ncont:
Thanks. I will give it a test now.

ncont:
Could you add a readme section? Especially about the lagacy mode and I like the table in description of this pr.

ncont:
Regarding 

```
It would be nice to keep it this way for those who update but turn legacy mode off by defaults for new installs.
```

I share your concern about upset users. Changing gesture behavior is pretty invasive. Especially since this shall just be a minor version update. 
Let's do it the following way: we leave the legacy mode in for the next feature releases and disable it by default for the next major release. If you want, you can add an issue to keep track of it and I will assign it to the next milestone. 

Deal?

ncont:
Hey. I am sorry that i dont find the time to look more into the issue, but here is a claude review after i was not able to build it: 


### 1. Doesn't compile
`about_screen.cpp` has a new gesture-summary table added *above* the old one instead of replacing it — two `String rows[...]` declared in the same scope, plus six lang macros (`D_ABOUT_GESTURE_CLICK`, `D_ACTION_NEXT_LABEL`, …) and an `actionLabel()` helper that are referenced but never defined anywhere.

Verified with `pio run -e wireless-paper-v1_2-en`: ~10 compile errors, all in this one file. Looks like a WIP block that was meant to replace the old one but got left alongside it.

### 2. Saving *any* other settings form silently turns legacy mode off
`web/settings.cpp::handleSettingsPost()`: the `legacy_cont` checkbox handling sits *outside* the `if (server.hasArg("btnS") || ...)` guard that protects the button bindings. `legacy_cont` only appears in the POST body when the Buttons form is submitted (same reasoning the existing code comment already gives for `btnL`/`btnXL`/`btnCH`). So saving, say, a font-size change — a different `<form>` — omits `legacy_cont`, and the code falls into the `else` branch and calls `setLegacyControls(false)` unconditionally. Every user who tweaks an unrelated setting gets bumped into remapped mode without touching the Buttons card.

**Suggested direction:** move that block inside the same `hasArg` guard used for the button fields.

### 3. Inverted condition breaks navigation in `list_screen.cpp`
New code: `if (!Gestures::isNonLegacyAction(e, ACTION_PREV)) { …decrement…; return; }`. `isNonLegacyAction` always returns `false` while legacy mode is on (the default), so `!false` is always `true` — this block fires and returns for *every* event that isn't the Short/NEXT case, before the Long ("mark done") and Triple ("go to library") handlers further down ever run.

Compare to `library_screen.cpp` and `apps_screen.cpp`, which add the same "previous item" feature with `if (Gestures::isNonLegacyAction(e, ACTION_PREV))` — no `!`. Looks like a stray negation copied into one file.

**Suggested direction:** dropping the `!` should restore the existing Long/Triple behavior while still adding PREV.

### 4. Uninitialized stack read in `drawBatteryTopRight`
`battery.cpp`: `char extendedInfo[11];` is only populated inside `if (extended) { snprintf(...) }`, but the `gfx.fillRect(..., u8g2.getUTF8Width(extendedInfo) + ..., ...)` call right below it is *not* inside that `if` — it runs on every call, including the plain `drawBatteryTopRight()` used on nearly every screen (`widgets.cpp:68`). That's a read of an uninitialized, not-necessarily-null-terminated buffer — undefined behavior, could size the clear-rect wrong or worse.

**Suggested direction:** guard the `fillRect` width computation the same way the `xIcon` computation already is.

### 5. `ButtonAction` enum was renumbered, but values are persisted as raw ints in NVS with no migration
Old: `BOOKMARK=1, LOCK=2, MENU=3, ROTATE=4`. New: `NEXT=1, PREV=2, MENU=3, LOCK=4, HOME=5, BOOKMARK=6, ROTATE=7`. `loadSettings()` just does `clamp(prefs.getInt(key, default))` — no version check. Anyone who previously customized `cfg_btnL`/`cfg_btnXL`/`cfg_btnCH` (e.g. stored `1` meaning old `ACTION_BOOKMARK`) will silently get reinterpreted as `ACTION_NEXT` after upgrading — a quiet behavior change for exactly the upgrading users the PR says legacy-mode-by-default is meant to protect.

**Suggested direction:** ask the author whether existing action ordinals can be kept stable and new ones appended at the end instead.

## Smaller / code-smell items (not blocking)

- Dead code left commented-out instead of deleted: the old Long/VeryLong/ClickHold dispatch block in `reader_screen.cpp` (with a now-stale "pre-PR-#15" comment), a redundant commented line in `resolveLegacyAction`, and `// Gestures::setLegacyControls(true);` in `settings.cpp`.
- `library_screen.cpp::onButton`: the `ACTION_LOCK` non-legacy branch returns immediately, but `ACTION_ROTATE`/`ACTION_HOME` fall through several unrelated `if`s and only stop because a guard clause further down happens to catch it — works, but fragile against future edits.
- Inconsistent formatting throughout the new code (misaligned `s_short`/`s_double`/... declarations, stray blank lines) — cosmetic, but the PR touches a lot of files and it'd read easier cleaned up.

gamberoillecito:
@ncont hi, I'm sorry you couldn't build it. I haven't tried honestly and simply trusted pio check.

gamberoillecito:
@ncont I fixed the build issue (I will also address the other ones that you/claude pointed out since they seem valid). I cannot understand why the check have failed. Running `pio check -e wireless-paper-v1_2-en --fail-on-defect low` locally passes all tests.

ncont:
> @ncont I fixed the build issue (I will also address the other ones that you/claude pointed out since they seem valid). I cannot understand why the check have failed. Running `pio check -e wireless-paper-v1_2-en --fail-on-defect low` locally passes all tests. 
> 

Thanks. Pio ckeck is "just" static code review. It finds a lot but can not resolve everything. I run pio ceck and then a build without the target 'upload'.

The CI needs to be improved to give you earlier feedback. One point of the endless to do list ;)

gamberoillecito:
> > @ncont I fixed the build issue (I will also address the other ones that you/claude pointed out since they seem valid). I cannot understand why the check have failed. Running `pio check -e wireless-paper-v1_2-en --fail-on-defect low` locally passes all tests.
> 
> Thanks. Pio ckeck is "just" static code review. It finds a lot but can not resolve everything. I run pio ceck and then a build without the target 'upload'.
> 
> The CI needs to be improved to give you earlier feedback. One point of the endless to do list ;)

Ok, what should I do then? 
Moreover, should I push the fixes to cluade's comments? I fixed everything except for 5, which I don't know how to handle

ncont:
> > > @ncont I fixed the build issue (I will also address the other ones that you/claude pointed out since they seem valid). I cannot understand why the check have failed. Running `pio check -e wireless-paper-v1_2-en --fail-on-defect low` locally passes all tests.
> > 
> > Thanks. Pio ckeck is "just" static code review. It finds a lot but can not resolve everything. I run pio ceck and then a build without the target 'upload'.
> > 
> > The CI needs to be improved to give you earlier feedback. One point of the endless to do list ;)
> 
> Ok, what should I do then? 
> Moreover, should I push the fixes to cluade's comments? I fixed everything except for 5, which I don't know how to handle

Push the fixes and make sure it could be deployed to a Pala.

To point 5: I should have removed it. Since not a lot have used the feature, there is no need for backwards compatibility

gamberoillecito:
Hi @ncont , I fixed the previously mentioned issues in the second to last commit but I think I messed up when solving the merge conflict in my last one. I'm having a hard time figuring this stuff out, can you solve it easily?

ncont:
> Hi @ncont , I fixed the previously mentioned issues in the second to last commit but I think I messed up when solving the merge conflict in my last one. I'm having a hard time figuring this stuff out, can you solve it easily?

I can try at least.

ncont:
@gamberoillecito I resolced and did some minor fixes. could you check again if this is ok for you and all works as expected ?

gamberoillecito:
> @gamberoillecito I resolced and did some minor fixes. could you check again if this is ok for you and all works as expected ?

Flashed it and everything seems to work! Thank you for the help!

ncont (approved):

Issue #70, opened 2026-05-28
Show loading icon for time consuming operations


gamberoillecito:
Certain operations (e.g. changing the status bar style) lock the device for several seconds making it unresponsive. It would be a great improvement for the user experience to show a loading icon to inform that something is going on. A user on discord reports doing something similar for a set of circumstances (e.g. loading books): https://discord.com/channels/1502006584639881256/1503249666936930324/1503249666936930324.

I haven't checked the code yet but I think a pair of functions `loadingStart()` `loadingFinish()` which respectively draw and cancel a loading icon somewhere on the screen could be enough for cases in which a given operation is known to be time consuming. For example:
```
loadingStart()
timeConsumingOperation()
loadingEnd()
```

gamberoillecito:
I will close this since #119 addresses the main (only?) case in which a loading indicator would be necessary

PR #116, opened 2026-06-30
Add an option to support half height paragraph breaks.

VeniVidiLudi:
Currently a lot of screen real estate is given over to paragraph breaks. This change adds an option for paragraph breaks to only span half a line instead of a full line. This means we can get more lines of useful text on the screen while still making paragraphs clear.

This required changing the screen layout calculation to rely on a pixel budget rather than just counting lines.  Changing the setting also invalidates the on disk page cache.

Other than that, the implementation mostly followed the same paths as the similar "bionic reading" feature.

NOTE: I do not speak Spanish and I used automated translation tools for the UI text.

### DIFF
--- a/Pala_One_2_1/src/pure/paginator.h
+++ b/Pala_One_2_1/src/pure/paginator.h
@@ -13,6 +13,8 @@ struct LayoutMetrics {
   int lineH = 0;
   int maxWidth = 0;
   int maxLines = 1;
+  // Height to be used for a paragraph gap. Either lineH or lineH/2.
+  int paragraphGapH = 0;
 };
 
 // Measure the rendered width (in pixels) of a UTF-8 string under the layout's
--- a/Pala_One_2_1/src/storage/page_cache.cpp
+++ b/Pala_One_2_1/src/storage/page_cache.cpp
@@ -43,12 +43,12 @@ static constexpr size_t kHeaderBytes =
 // statusbarReserve in pixels (currently 0/1/STATUS_H) gets to stretch out in the top byte.
 // Bits 20-23 are currently spare.
 static uint32_t encodeLayoutVersion(const PageCacheLayout& layout) {
-  return ((uint32_t)(layout.bodySize         & 0x0F))
-       | ((uint32_t)(layout.lineGap          & 0x0F) << 8)
+  return ((uint32_t)(layout.bodySize         & 0xFF))
+       | ((uint32_t)(layout.lineGap          & 0xFF) << 8)
        | ((uint32_t)(layout.family           & 0x03) << 16)
        | ((uint32_t)(layout.bionic           & 0x01) << 18)
        | ((uint32_t)(layout.halfGaps         & 0x01) << 19)
-       | ((uint32_t)(layout.statusbarReserve & 0x0F) << 24);
+       | ((uint32_t)(layout.statusbarReserve & 0xFF) << 24);
 }
 
 static String pageCachePathForBook(const String& path) {
--- a/Pala_One_2_1/src/storage/page_cache.cpp
+++ b/Pala_One_2_1/src/storage/page_cache.cpp
@@ -47,7 +47,7 @@ static uint32_t encodeLayoutVersion(const PageCacheLayout& layout) {
        | ((uint32_t)(layout.lineGap          & 0x0F) << 8)
        | ((uint32_t)(layout.family           & 0x03) << 16)
        | ((uint32_t)(layout.bionic           & 0x01) << 18)
-       | ((uint32_t)(layout.bionic           & 0x01) << 19)
+       | ((uint32_t)(layout.halfGaps         & 0x01) << 19)
        | ((uint32_t)(layout.statusbarReserve & 0x0F) << 24);
 }
 
--- a/Pala_One_2_1/src/storage/page_cache.cpp
+++ b/Pala_One_2_1/src/storage/page_cache.cpp
@@ -21,12 +21,14 @@
 //                 must rebuild when Statusbar::setMode toggles between
 //                 Full / Minimal / Hidden because each has a different
 //                 reserve height and therefore a different maxLines
+//    0x50434F4A — added the half-height-paragraph-gaps flag; toggling it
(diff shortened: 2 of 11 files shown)


gamberoillecito:
Hello, I really like the idea of this PR (it may also be set as default IMO). I just installed it and verified that it is working (see attached pictures).
The only issue I found so far is that, after applying the settings and going back to the book, the device freezes while the book pagination is re-adjusted. I believe this is not an issue purely of this PR (see #70) but it can be quite unsettling for those who don't know about the technical details if the devices freezes completely.
signal-2026-07-03-11-58-17-183
signal-2026-07-03-11-58-17-183_002

VeniVidiLudi:
I think the most likely culprit is the invalidation of the page cache and the need to rebuild it when next opening the book. I'll do some quick tests to confirm, leave it with me.

If it is that, I guess the simplest fix would be to have an interstitial page when rebuilding the cache to let the user know something is happening.

VeniVidiLudi:
Well, that was quite a rabbit hole. It's definitely the page cache and I have some thoughts on broader improvements: #118 . I also stumbled across a weird bug when changing the status bar visibility #117.

Anyway, I'm going to create a new PR to add the small interstitial page to at least let the user know the reason for the delay.

I leave it to you whether this PR should block on that one or not.

Pala_One_2_1/src/storage/page_cache.cpp
```diff
@@ -35,16 +37,18 @@ static constexpr size_t kHeaderBytes =
...
+// growth; and, bionic ∈ {0,1} and halfGaps ∈ {0,1} get a bit each.
+// statusbarReserve in pixels (currently 0/1/STATUS_H) gets to stretch out in the top byte.
+// Bits 20-23 are currently spare.
 static uint32_t encodeLayoutVersion(const PageCacheLayout& layout) {
-  return ((uint32_t)(layout.bodySize         & 0xFF))
-       | ((uint32_t)(layout.lineGap          & 0xFF) << 8)
-       | ((uint32_t)(layout.family           & 0x0F) << 16)
-       | ((uint32_t)(layout.bionic           & 0x0F) << 20)
-       | ((uint32_t)(layout.statusbarReserve & 0xFF) << 24);
+  return ((uint32_t)(layout.bodySize         & 0x0F))
```
ncont: The space for lineGap and statusbarReserve are smaller now. Which is fine. I doubt there will be ever a font size grater then 14 and the status bar is already big enough. However, a comment for `STATUS_H` would be nice that the max is 15px now. Maybe someone would like to have a double line status bar or so. 
Or, we could leave it at the value before. There is still enough headroom in uint32.

Pala_One_2_1/src/pure/paginator.h
ncont, line 25: The comment above `LineCallback` still says `buf` is always "NUL-terminated and trailing-spaces-trimmed" with no mention it can now be `nullptr`. The two existing call sites in `text.cpp` do check for it, but only because the author remembered to — the header itself won't warn the next caller.

**Suggested direction:** add a line noting `buf == nullptr` signals a paragraph-gap event.

ncont (commented):
`pure/paginator.cpp` is the host-tested module this PR most heavily reworks (line-count → pixel-budget tracking), but `test/test_paginator.cpp` isn't touched at all. The `LineCallback` contract also changed — `onLine` can now fire with `buf == nullptr` for a paragraph gap — but the test file's `linesOf()` helper does `lines.push_back(String(buf))` unconditionally at `test/test_paginator.cpp:30`, so a natural test for a `\n\n` case using the existing helper would pass `nullptr` into `String(...)`.

**Suggested direction:** add a test covering a mid-page paragraph break and one at the very top of a page (the new skip-leading-blank case), and have the helper handle/guard the `nullptr` sentinel.

Pala_One_2_1/src/storage/page_cache.cpp
VeniVidiLudi: Just to be clear, lineGap and statusbarReserve stay at a byte each - they have not changed. What I have squeezed is _family_ down from four bits to two and _bionic_ down from four bits to one. Bionic is a boolean, so that seems fine. Family is more of a judgement call, do we expect more than four fonts in the future?

We have four bits left over that things can expand into, obviously if we knew which fields we expect to expand, we could give them the headroom now and then it wont require a magic byte change when it eventually happens. Otherwise I'd suggest we just wait and see.

Do you want to proactively give family more headroom to grow or just keep things as is for now?

Pala_One_2_1/src/pure/paginator.h
VeniVidiLudi: Reflecting on this code, I actually don't know why I leant on nullptr here in the first place. I can't see any good reason why it has to be indicated by something particularly special (as opposed to just an empty line) and I really don't see why it has to be something as potentially disruptive as nullptr in particular. If I had a reason in the moment, it's been lost to the ravages of time.

I've changed the code to just use an empty string instead of nullptr and I've changed the text.cpp file to just key off the len parameter. If it does turn out there's some corner case this doesn't cover later, I think it's better to try and fix it without using nullptr anyway.

VeniVidiLudi:
I've added tests and addressed other comments as requested. PTAL.

Pala_One_2_1/src/storage/page_cache.cpp
ncont: But you changed from 0xFF to 0x0F for status reserve. Or do we look at different changes maybe.

VeniVidiLudi: Gosh, well spotted and that's totally my mistake. I had at one point intended to compress things much more aggressively, but decided against it. I did fix up the comments and the shift amounts, but I must have copy/pasted the wrong bitmask back into place.

Sorry about that, I've fixed it now.

ncont: No problem at all. 

And yes the datatypes and their usage are very... Suboptimal rn in the code overall. At some point it will be a good idea to reduce the footprint of the FW. But maybe not now where a lot is changing. 

Thanks for the fix!

Issue #118, opened 2026-07-06
Rebuilding the page cache causes long freezes


VeniVidiLudi:
### What’s The Issue?
The Pala One maintains a “page cache” that maps Page Numbers to byte offsets within the file. Changing the layout settings on the Pala One (e.g. font size, line spacing) changes how pages will map by changing the amount of text that fits on each individual page. As such, any change in the layout settings (with the exception of the status bar, which we’ll discuss later) will result in the page cache for all books being marked as invalid.

When the page cache is invalid, it has to be rebuilt the next time the book is opened. This involves sequentially reading the file from start until the current reading position and recalculating each page along the way. This sequential nature means that it will grow linearly the further you are into the book. This can manifest as “freezes” lasting several seconds when opening a long book where the reading position is deep into it.

This same effect can happen when using the skip ahead, although there the time taken is proportional to the difference between the current reading position and the position being skipped to.

### How Can I Recreate?
1. Upload a reasonably long book to your device.
2. Use the Web UI to skip far into the book.
3. Open the book on your device.
4. Notice the long pause when opening.
5. Use the Web UI to change something that affects layout (e.g. font size or line gap).
6. Open the book again on your device.
7. Notice another long pause when opening.

### What About The Status Bar? You Said That Didn’t Invalidate The Cache, How?
The status bar was handled a little differently because it is a setting that is changed during reading mode where the pain of a long freeze was more obvious. An optimisation was put in place to only invalidate the cache from the current point forward (which is probably empty anyway in the most common usage). The cache is still used for backwards navigation or jumps, and things being slightly different than what you would get if you started at the beginning and moved forward is just accepted as a reasonable tradeoff.

So, the obvious question might arise as to why we can’t just do the same for other settings and the reality is there’s a nasty bug hidden in that optimisation. You can read about it in issue #117 but the short version is that if you use this optimisation, some of the content become inaccessible. For the status bar this was probably never a huge issue as many font sizes were somewhat immune and you would generally be missing at most one line. For other layout changes the number of lines impacted could be much larger and you could wind up with half the content being unreachable.

### What’s The Fix?
I guess the quickest easiest fix is to at least let the user know something is happening during the rebuild. This is probably also the catalyst for issue #70. I’d propose a simple splash screen with either a percentage completion or a progress bar.

However, I honestly feel that this is a bit of a band-aid on what is a much larger issue: the page cache should probably be done away with.

[...]

PR #119, opened 2026-07-06
Show progress of long page index rebuilds.

VeniVidiLudi:
Rebuilding the page index can take a while especially if you're deep into a book. At the moment this presents to the user as just a long freeze with no explanation.

This change adds a progress interstitial  when index rebuilds take longer than some threshold (currently 500ms). Progress is shown by a bar that fills as progress is made.

This will address issue #70 and is the beginning of adressing #118. It was also brough up during PR #116.

### DIFF
--- a/Pala_One_2_1/src/ui/reader.cpp
+++ b/Pala_One_2_1/src/ui/reader.cpp
@@ -168,10 +168,15 @@ static void ensureOffsetsUpTo(int targetPage) {
 // rebuilds.
 static const uint32_t kRebuildSplashDelayMs  = 500;
 // Controls how long between progress updates.
-// Screen updates take a while, so there's no point in updating constantly.
+// Screen updates take a while, so updating too frequently will actually increase
+// the time a rebuild can take substantially.
 static const uint32_t kRebuildSplashRedrawMs = 700;
 
 static void drawCacheRebuildProgress(int pct, bool firstFrame) {
+  const int barX = MARGIN_X;
+  const int barW = SCREEN_W - 2 * MARGIN_X;
+  const int barY = 74;
+  const int barH = 10;
   if (firstFrame) {
     prepareMenuFrame();
     Font::useBody();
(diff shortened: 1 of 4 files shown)


gamberoillecito:
I love this! Would it make sense/be feasible to separate the progress bar to it's own "component" to be used in other cases such as installing the new firmware, searching for updates and so on?

VeniVidiLudi:
The drawing of the progress bar itslef is mostly contained to a single function that wouldn't be too hard to repurpose. The title and body text are hard coded and would need to be parameterised but that's trivial.

I don't have much familiarity with the firmware update stuff yet, so I can't say for sure whether you'd be able to get a reasonable display of progress there (it needs to have some defined end to measure progress against and it needs to not be a single blocking call). Assuming you can, no reason it couldn't be pulled out of reader to somewhere common and reused from there. If it is a single blocking call, you could try and hand the progress drawing off to a separate thread (I was surprised to learn the Heltec board is actually a dual-core device) but any multithreading makes things more complicated.

With that said, I would caution something about this progress bar in general: due to the slow nature of screen updates on e-ink and the fact that display.update() is blocking, adding a progress bar actually increases the time a task will take. I think this is fine for page index rebuilds as those typically wont be too long (many will finish before we even show the page, most will probably be 1-2 screen draws). But it might be worth considering the impact for longer tasks and tune down the update frequency accordingly.

So, feasible? Yes. Makes sense? Maybe, it depends. I would personally wait on generifying it until we're hooking it up to a second callsite.

Pala_One_2_1/src/ui/reader.cpp
```diff
@@ -163,13 +163,70 @@ static void ensureOffsetsUpTo(int targetPage) {
...
+static void drawCacheRebuildProgress(int pct, bool firstFrame) {
+  const int barX = MARGIN_X;
+  const int barW = SCREEN_W - 2 * MARGIN_X;
+  const int barY = 74;
+  const int barH = 10;
+  if (firstFrame) {
+    prepareMenuFrame();
+    Font::useBody();
+    int ascent = u8g2.getFontAscent();
+    int lineH  = (ascent - u8g2.getFontDescent()) + Font::currentLineGap() + 1;
```
ncont: Linter is complaining that the variables are set but never used.

VeniVidiLudi: Fixed

VeniVidiLudi:
I think this should be good to go now.

ncont (approved):

PR #122, opened 2026-07-07
Allow WiFi settings to be managed from the Web UI.

### LINKED ISSUE #121: Make WiFi settings editable in the Web UI
At the moment he WiFi settings can only be set via USB. We should make them also editable via the Web UI so that we can manage them more easily, especially on the go.

VeniVidiLudi:
Adds a card at the end of the existing settings page that provides another path to setting/clearing wifi settings for the device than just USB provisioning. Very useful when you're on the road.

The password is hidden by default but can be shown by clicking on a small SVG button made to look like an eye. I could have used unicode characters to make life simpler, but they wouldn't have followed the light/dark theme and would look different depending on OS, so this felt like a better fit.

Closes #121.

Screenshot 2026-07-07 231751
Screenshot 2026-07-07 231806
Screenshot 2026-07-07 232450
Screenshot 2026-07-07 232501

### DIFF
--- a/Pala_One_2_1/src/web/settings.cpp
+++ b/Pala_One_2_1/src/web/settings.cpp
@@ -348,12 +348,12 @@ static void handleSettingsPost() {
   // POST won't accidentally wipe the stored network via absent fields.
   // A blank SSID forgets the stored network.
   if (server.hasArg("wifi_form")) {
-    String ssid = server.hasArg("wifi_ssid") ? server.arg("wifi_ssid") : "";
+    String ssid = server.hasArg("wssid") ? server.arg("wssid") : "";
     ssid.trim();
     if (ssid.length() == 0) {
       WifiCreds::clear();
     } else {
-      WifiCreds::save(ssid, server.hasArg("wifi_pass") ? server.arg("wifi_pass") : "");
+      WifiCreds::save(ssid, server.hasArg("wpass") ? server.arg("wpass") : "");
     }
   }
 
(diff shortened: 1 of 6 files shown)


ncont:
Good feature! I would like to perform some tests. But since this feature was requested by some other users as well, it will be merged as soon as I can.

ncont:
Would be great if you could add a readme section about this feature and how to use.

Pala_One_2_1/src/web/settings.cpp
```diff
@@ -188,6 +189,39 @@ static void handleSettings() {
...
+    "<div class='grid cols-2'>"
+    "<div><label for='wssid'>" D_WEB_WIFI_SSID_LABEL "</label>"
+    "<input type='text' id='wssid' name='wssid' maxlength='32' placeholder='" D_WEB_WIFI_SSID_PLACEHOLDER "' value='";
+  out += htmlAttrEscape(WifiCreds::ssid().c_str());
+  out +=
+    "'></div>"
+    "<div><label for='wpass'>" D_WEB_WIFI_PASSWORD_LABEL "</label>"
+    "<div class='pwwrap'>"
+    "<input type='password' id='wpass' name='wpass' maxlength='64' placeholder='" D_WEB_WIFI_PASSWORD_PLACEHOLDER "' value='";
+  out += htmlAttrEscape(WifiCreds::pass().c_str());
```
ncont: For me this is a security concern. I now of users who have their details as screensaver in case they loose the device. It would be pretty bad if you not only have the details but the Wifi Password in clear text as a bonus. 

So, i would use the field as a one way route only. password write but never read.

VeniVidiLudi: Nice catch. I've replaced the wifi password with a placeholder when set. Getting this placeholder back on form submission is interpreted as "keep existing password".

I've left the hiding/showing of the field contents in place to cover the "someone looking over your shoulder" case while entering new details.

image

VeniVidiLudi:
Added README section and addressed security concern. PTAL.

ncont (approved):

Issue #120, opened 2026-07-06
Allow switching layout settings on device.


VeniVidiLudi:
Most of the layout settings are simple toggles or selections from a short list. We should be able to change these on-device without having to use the Web UI to do so.

gamberoillecito:
I would personally go the other way and remove also the setting for the progress bar as suggested in #112. At least for my use case I don't think the added complexity makes it worth it: I change settings once a month or less (once I found something that works I don't think there is much need for messing with it). It may be worth a discussion both with maintainers and users to see what's best

VeniVidiLudi:
For "added complexity" do you mean mostly interactions with the page index? Because I think I've made my thoughts clear on that in #118 (TL;DR: I think we should do away with page numbers and remove the index altogether).

For me, I found the process of dialing in the settings at first a little laborious (pull out phone, do the pala wifi network dance, update settings, disconnect, try them out, lather, rinse, repeat). I also find that while tiny font is my preferred reading size - just due to reducing number of page turns - it can be a little tough to read in poor lighting conditions and I sometimes wish I could just quickly bump the size up without having to resort to pulling out my phone to do so (trying to not using my phone was the main reason I built a Pala One in the first place).

I do think the WiFi settings should be a superset of the on-device settings and there's a lot of things that don't make sense to try and do on device, but for simple toggles and switches, it would be nice to be able to just use the device in isolation, especially when you're on the road.

Obviously if others feel strongly that this is a bad move I wont bother.

Issue #61, opened 2026-05-26
Customizing main menu


tomascorreag:
It would be nice to be able to enable/disable/reorder the items in the main menu, from the Web UI. I personally enjoy the minimalism of the original firmware, and perhaps even removing the Bookmarks and Device screens would be cool too.

ncont:
Yes. I noticed that there is a not commonly used strategy. When there are no List entries, there is no menu entry. 
But if there are apps or Bookmarks - the entry is always there. Maybe it is a good idea to remove items from the Menu, if they have no entry.

gamberoillecito:
@ncont I'd like to be assigned to this. Let me know if it is fine and whether to start working on it or wait a bit for the current changes to settle down

ncont:
Please! It is all yours. The MR for #120 is waiting for this Mr. 
Would be very nice if you could implement it!

VeniVidiLudi:
I really like the customization options, and I wonder if, off the back of that, we should take things back to their minimalist roots and pare the default set down to the bare bones. This keeps the initial experience clean but still allows people to turn things on as they get comfortable with the device. My vote would be to just have Bookmarks, Device and Upload, but I could even see the argument for dropping Bookmarks, tbh.


Interested to hear what others think.

Higgs2209:
> I really like the customization options, and I wonder if, off the back of that, we should take things back to their minimalist roots and pare the default set down to the bare bones. This keeps the initial experience clean but still allows people to turn things on as they get comfortable with the device. My vote would be to just have Bookmarks, Device and Upload, but I could even see the argument for dropping Bookmarks, tbh.
> 
> Interested to hear what others think.

I would personally vote to have just Device, Upload and Layout settings. I haven't seen much discussion of people using 'Bookmarks' on the discord.

gamberoillecito:
@VeniVidiLudi @Higgs2209 thanks for your comments. I agree that it would be nice to create a "ready to use" interface for **new** users, but how would you deal with existing users? I think it would be quite annoying to see my layout changed after an update. Is there a way to distinguish between an update and a new install so that, in the first case, the main menu remains as it was and, in the second, a minimal set of options is displayed?

Moreover, I would argue that it would be better to present new users with the full list of menus and let them decide which to remove, since otherwise I believe it would be difficult to discover them

PR #123, opened 2026-07-08
Add on-device Layout Settings menu

VeniVidiLudi:
Add an on device "Layout Settings" menu to the main screen. This allows changing page layout settings without resorting to the Web UI.

Satisfies issue #120


ncont:
Hey, 
thanks for your contribution! 

Based on the community feedback on Discord, the trend is going the "less is more" approach. Meaning to offload settings and so on off the device and push it to the Web UI. 

I dont want to say no to this PR as i can imagine that some will like it. As a compromise, i would wait until the "confiurable menu" is in place (Dont know if there is a Ticket for it. But some talked about it).

VeniVidiLudi:
Yeah, saw that discussion, can't argue with the complaint and, by extension, agree that merging this now would be the wrong thing to do. Personally, I've been commenting out things like Apps and Statistics on my own devices for a while now.

FWIW, despite being somewhat sensitive to bloat, after having patched this into my own devices, I have found it to be a useful thing to have ready access too. So, I do think it's something that might merit merging in once the dust has settled. Perhaps off by default though.

I'm not going to have much time for dev work for the next few weeks, but if no one else has addressed the clutter issue before middle of August or so I might take a swing at it.

ncont:
> Yeah, saw that discussion, can't argue with the complaint and, by extension, agree that merging this now would be the wrong thing to do. Personally, I've been commenting out things like Apps and Statistics on my own devices for a while now.
> 
> FWIW, despite being somewhat sensitive to bloat, after having patched this into my own devices, I have found it to be a useful thing to have ready access too. So, I do think it's something that might merit merging in once the dust has settled. Perhaps off by default though.
> 
> I'm not going to have much time for dev work for the next few weeks, but if no one else has addressed the clutter issue before middle of August or so I might take a swing at it.

Sounds like a plan. The issue is #61, just to keep track.

ncont:
Hey. #134 is merged and now you could rebase and make the menu entry customizable. :)

VeniVidiLudi:
Okay, I've merged this with the configurable menu stuff, so it should be ready for review.

Note, I've left this menu off by default, I'm actually going to propose we revist what is in on by default on the device to keep the initial experience as smooth as possible. I've added a note to #61 to prompt discussion.

Pala_One_2_1/src/ui/screens/library_screen.cpp
```diff
@@ -234,59 +236,47 @@ void LibraryScreen::onButton(const ButtonEvent& e) {
...
-  if (sel.type == LIB_ENTRY_BOOK) {
-    if (openBookByIndex(sel.ref)) {
-      nextScreen = &g_readerScreen;
-    } else {
-      drawCenter(D_LIBRARY_OPEN_FAILED, D_LIBRARY_TRY_UPLOAD);
+  switch(sel.type) {
+    case LIB_ENTRY_ABOUT:
+      nextScreen = &g_aboutScreen;
+      return;
+	case LIB_ENTRY_APPS:
```
Higgs2209, line 243: Case labels are tab-indented here (243–277) while the bodies are spaces. Rest of the tree is 2-space throughout.

Pala_One_2_1/src/pure/library_nav.h
```diff
@@ -32,13 +32,14 @@ struct Catalog {
 };
 
 enum LibraryEntryType {
```
Higgs2209, line 34: These ordinals are persisted. `library_menu_order.cpp:66 `serialises them into `cfg_lib_order` and casts the ints straight back on load, so the declaration order here is a wire format.
Once the next v* tag goes out these numbers are permanent and a reorder stops being fixable.

It does bite the dev channel today. I ran the stock saved order `("2,4,5,3,6,8,7")` through this branch: it comes back as List, Bookmarks, Layout Settings, Update, Statistics, Connect, with Bookmarks and Apps dropped entirely — they land on `BOOK `and `FOLDER `and fail `isValidLibEntry`. Anyone testing off `/dev/` who's used the library-order form since late July gets that on next boot.

The alphabetical ordering can stay, it just needs the values fixed:
```
// Numeric values are persisted to NVS by LibraryMenuOrder (cfg_lib_order) and
// cast back on load, so they are a wire format. Declaration order below is
// cosmetic; the values are not. New entries take the next unused number.
enum LibraryEntryType {
  LIB_ENTRY_ABOUT      = 6,
  LIB_ENTRY_APPS       = 4,
  LIB_ENTRY_BOOK       = 1,
  LIB_ENTRY_BOOKMARKS  = 2,
  LIB_ENTRY_FOLDER     = 0,
  LIB_ENTRY_LIST       = 3,
  LIB_ENTRY_SETTINGS   = 9,
  LIB_ENTRY_STATISTICS = 5,
  LIB_ENTRY_UPDATE     = 7,
  LIB_ENTRY_UPLOAD     = 8
};

```

Could you add a case to `test_library_nav.cpp` locking the numbers down too? Otherwise the only thing protecting the format is that someone remembers it's a format.

Higgs2209 (changes requested):

gamberoillecito:
I think it could make sense to address #112 too here, what do you think?

VeniVidiLudi:
Folks, just wanted to give a quick sign of life update here. I haven't forgotten about this, but circumstances have prevented me from working on it. I hope to return to it within the next 2 weeks.

PR #125, opened 2026-07-09
Buffer file reads to improve pagination performance

VeniVidiLudi:
At the moment, every byte read from file round trips twice down the entire file handling stack: once to check for next byte availability; and again to actually retrieve the next byte. Even though there are buffers at the bottom of that stack, it's still a substantial cost to pay on a per byte basis.

This change adds a small (512 byte configurable) buffer wrapping all file reads to reduce the number of stack walks we do.

In experiments, this improved individual page rendering time by 50-75% which was especially noticeable when rebuilding page caches. Here's some quick data for one experiment reconstructing the page caches for two books using different buffer sizes:

Book | Vanilla | 4K Buffer | 2K Buffer | 512b Buffer | 256b Buffer
-- | -- | -- | -- | -- | --
DCC | 4944 ms | 1978 ms | 414 ms | 342 ms | 323 ms
long number | 31546 ms | 11058 ms | 5269 ms | 4854 ms | 4871 ms

Note the 256b buffer showed only marginal improvements than 512b and in a separate test performed a little worse when fed unicode inputs (presumably due to going over the buffer size more often) so I went with 512b as the default. Should also mention that that experiment was conducted with the progress bar change in #119 in place. This means that some of the time saving comes from fewer screen updates and not just faster pagination. This is especially true of the first book where it managed to get it under the delay for showing the progress bar at all.

This should go some way to helping to alleviate #118.

### DIFF
--- a/Pala_One_2_1/src/config.h
+++ b/Pala_One_2_1/src/config.h
@@ -110,6 +110,12 @@ static const float BAT_CHARGING_VOLTAGE = 4.2f;
 static const int FULL_REFRESH_EVERY_N_PAGES = 100;
 static const int MENU_FULL_REFRESH_EVERY = 60;
 
+// Buffer size for file read operations. Bigger is not necessarily
+// better as you might wind up just pulling in more data than you need
+// to fill a page. This was a value that worked well with both ASCII
+// unicode inputs.
+static const size_t FILE_STREAM_BUF_BYTES = 512;
+
 static const int MARGIN_X = 6;
 static const int TOP_PAD = 3;
 static const int BOT_PAD = 0;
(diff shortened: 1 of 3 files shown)


Higgs2209 (approved):

Higgs2209:
I am happy to merge this but I want to see if @ncont is still around as a second set of eyes on everything would be helpful!

ncont:
> I am happy to merge this but I want to see if @ncont is still around as a second set of eyes on everything would be helpful!

I am back next week. Will have a look as soon as possible

ncont (approved):

PR #126, opened 2026-07-09
Speed up line length computation

VeniVidiLudi:
This change alters how we compute the total line length as part of pagination. In the old system, we would build up a scratch with every token and then call "measure" to determine the width of the whole scratch each time to see if it had to be a new line. This is enormously wasteful and gets worse as lines get longer.

We've instead switched it to only track the portion of the current line since the last whitespace (which should act as a width anchor) and recompute the width of just that mutable tail based on the influence of the new token.

This is layout identical to the existing approach and so doesn't require a page cache invalidation.

Some crude experiments show a performance improvement of ~30% on page cache rebuilds:
Book | 512b Buffer | Tail Width Tracking
-- | -- | --
DCC | 551 ms | 350 ms
long number | 3425 ms | 2215 ms

It's worth noting while that's a pretty good improvement, it's taken on the Heltec board with just 250 pixel width. Because the line scan redundancy gets worse with line length, this was slowing page turns to a crawl on larger devices. I swear on the LilyGo 4.7" at 960px wide, it would have been quicker to render on an Etch-a-Sketch sometimes.

It's worth noting that this code is actually still very over-engineered. The u8g2 fonts used by the Pala One do not have any kerning tables so there's actually no need to recompute any widths since they'll never change based on adjacent tokens. For now, however, I propose we push forward with this as is and we're future proofed against someone adding a font with kerning at a later date.

I had Claude make me a little harness to check that this paginates identially to the old code and it was found to be completely identical across the 192 test samples. As such, no modification of the magic number is needed: existing page caches should work fine.

### DIFF
--- a/Pala_One_2_1/src/pure/paginator.cpp
+++ b/Pala_One_2_1/src/pure/paginator.cpp
@@ -71,8 +71,8 @@ uint32_t paginatePage(IReadStream& in,
     emit(line, lineLen);
     lineLen = 0;
     lineW = 0;
-	tailStart = 0;
-	tailW = 0;
+    tailStart = 0;
+    tailW = 0;
   };
 
   auto safeReturn = [&](uint32_t off) -> uint32_t {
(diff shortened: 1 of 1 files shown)


Higgs2209:
This is a good idea - I was hoping someone would implement something along these lines.

I will review this evening

Pala_One_2_1/src/pure/paginator.cpp
```diff
@@ -102,48 +126,60 @@ uint32_t paginatePage(IReadStream& in,
...
-    } else {
-      memcpy(line + lineLen, token, tokLen);
-      lineLen += tokLen;
+      return startLineWithToken(measure(token));
     }
+
+    // The token fits entirely, perhaps having modified the tail
+	// width. Regardless, candidateW is the new width of the line.
+	// The tail only moves forward on whitespace so for the moment
+	// the start remains the same and the width becoems combinedW.
```
ncont:
```suggestion
	// the start remains the same and the width becomes combinedW.
```

ncont (commented):
Looks good to me. Please fix the mixed tab/spaces indent and i can merge it.

ncont:
Hey, since the comments were very minor and you are just before your vacation, I fixed the issue. Hope this is OK for you.

ncont (approved):

PR #128, opened 2026-07-10
Forgiving snippet matching + fix byte offsets sent to /jumpoffset

ssmit523:
What was changed

Two changes to the find UI in `src/web/find.cpp` (embedded JS only, no
firmware-side logic touched):

1. **Forgiving matching.** The search query now gets the same typography
   folding that book uploads get on the device (smart quotes, en/em dashes,
   ellipsis, and non-breaking spaces folded to ASCII), and both the query and
   the book text are compared case-folded with whitespace runs collapsed.
   Before this, pasting a sentence from a phone or e-reader app usually
   failed to match: those apps produce curly quotes and em dashes, while the
   stored book text was normalized to ASCII at upload, and any line break in
   the pasted text broke the exact `indexOf` match. An index map converts
   normalized match positions back to raw text positions, so highlighting
   and jump targets stay exact.

2. **Bug fix: `/jumpoffset` was sent JS string indices, not byte offsets.**
   `hits[i].start` is a UTF-16 string index into the decoded text. For any
   book containing multi-byte UTF-8 characters (accented letters etc.), that
   number is smaller than the real byte offset, so "Jump to here" landed
   before the intended position - drifting further the deeper into the book
   the match is. Offsets are now computed with `TextEncoder` so the device
   receives true UTF-8 byte offsets. The match counter also displays the
   real byte offset.

### Board versions tested

- Heltec V1.2: [flashed and smoke-tested - shows up in webui and worked when finding place in a book I wanted to pick up on the Pala One]
- Heltec V1.1: not tested (change is board-independent: embedded JS only)

### How it was tested

- The normalization/matching/offset logic was extracted and run under Node
  with a 13-case test suite: smart quotes, em dash, ellipsis, NBSP, case
  folding, whitespace collapse, matches across newlines, multiple
  occurrences, empty queries, and byte-offset correctness verified against a
  byte-level scan of the UTF-8 encoded text. All passing.
- `src/web/find.cpp` compile-checked with the ESP32-S3 toolchain
  (arduino-esp32 3.3.10, BOARD_V1_2, LANG_EN) - no new warnings.
- On device: flashed firmware and then tested feature in the webui; worked as expected

[README_v2.1beta_feature_find_my_place.md](https://github.com/user-attachments/files/29870087/README_v2.1beta_feature_find_my_place.md)


Higgs2209 (approved):

ncont (approved):

tomascorreag:
@ncont I see PR 128 was merged into main but hasn't been back-merged into dev, is there a reason for it?

ncont:
> @ncont I see PR 128 was merged into main but hasn't been back-merged into dev, is there a reason for it?

Good spot. That's why you shouldn't do it while on vacation 🙃

I will check it again when I am back.

PR #134, opened 2026-07-19
Customizable menu

gamberoillecito:
This PR addresses #61 making the main menu of the reader customizable by the user from the web UI.
Most of the coding has been done by a LLM but I've reviewed and tested it finding no issues for the moment.

Notes:
1. The new features seems to be robust to user errors and correctly deal with cases in which user removes the "Upload" entry, which is mandatory not to lock them out of the web ui.
2. The system silently handles and prevents duplicate entries in the menu
3. The system "compacts" the user-edited list of entries as not to leave "holes" in the new list.
4. The only other change in behaviour is the fact that the "list" entry now shows up even if empty, unless manually hidden by the user in the web ui.


Pala_One_2_1/src/storage/library_menu_order.cpp
```diff
@@ -0,0 +1,182 @@
...
+  s_count = normalizedCount;
+}
+
+static void persist() {
+  String encoded;
+  for (int i = 0; i < s_count; i++) {
+    if (i > 0) encoded += ',';
+    encoded += (int)s_order[i];
+  }
+  prefs.putString(kNvsKey, encoded);
```
ncont, line 75: Please use the [KV_Store](https://github.com/PaulLagier/pala-one-firmware/blob/main/Pala_One_2_1/src/storage/kv_store.h) as a abstraction layer for the settings like https://github.com/PaulLagier/pala-one-firmware/blob/0974ec383ca88eb9df34cee98383d19004ea4efa/Pala_One_2_1/src/storage/book_metadata.h#L4

This will narrow down the dep to the arduino ide and will make later refactoring and testing easier

Pala_One_2_1/src/storage/library_menu_order.cpp
```diff
@@ -0,0 +1,182 @@
...
+  normalizeOrder();
+  persist();
+}
+
+void resetToDefaults() {
+  prefs.remove(kNvsKey);
+  copyDefault();
+}
+
+}  // namespace LibraryMenuOrder
```
ncont: Missing new line at end of file.

Pala_One_2_1/src/storage/library_menu_order.h
```diff
@@ -0,0 +1,21 @@
...
+int count();
+LibraryEntryType entryAt(int index);
+int copyEntries(LibraryEntryType* out, int outCap);
+
+void setEntries(const LibraryEntryType* entries, int entryCount);
+void resetToDefaults();
+
+}  // namespace LibraryMenuOrder
+
+#endif  // PALA_STORAGE_LIBRARY_MENU_ORDER_H
```
ncont: Missing new line at end of file.

Pala_One_2_1/src/web/settings.cpp
```diff
@@ -71,6 +72,105 @@ static void appendActionSelect(String& out, const char* nameId, const char* labe
...
+    case LIB_ENTRY_APPS:      return D_MENU_APPS;
+    case LIB_ENTRY_STATISTICS: return D_MENU_STATISTICS;
+    case LIB_ENTRY_ABOUT:     return D_MENU_DEVICE;
+    case LIB_ENTRY_UPDATE:    return D_MENU_UPDATE;
+    case LIB_ENTRY_UPLOAD:    return D_MENU_UPLOAD;
+    default:                  return D_WEB_LIBRARY_ORDER_HIDDEN;
+  }
+}
+
+static bool isLibraryMenuEntryValue(int value) {
```
ncont: i feel that the helper functions for the lib entries are pretty far away from the actual enum. This will make adding a new lib entry bit hard as you have to know where they are. 
I think it would be good to consolidate them into one file and use them as function in the others ->One place to handle lib entries.

I found similar functions in : Pala_One_2_1/src/storage/library_menu_order.cpp:26-39 and Pala_One_2_1/src/ui/screens/library_screen.cpp:110-139.

Pala_One_2_1/src/ui/screens/library_screen.cpp
```diff
@@ -5,7 +5,8 @@
 #include "src/pure/library_nav.h"         // buildLibraryEntries
 #include "src/pure/paths.h"               // folderLeafLabel, bookLeafLabel
 #include "src/storage/library.h"
-#include "src/storage/list_items.h"       // listHasVisibleItems
+#include "src/storage/library_menu_order.h"
+#include "src/storage/list_items.h"       // g_list, listHasVisibleItems
```
ncont: Stale comment. `listHasVisibleItems` is removed and not included anymore.

Pala_One_2_1/src/storage/library_menu_order.cpp
gamberoillecito: Hi, thank you for all the comments, I don't understand what I should do in this case though. I followed the same approach I used in `screen_settings.cpp`, are these two things different? In any case I'm having some troubles figuring out what to change to comply

Pala_One_2_1/src/web/settings.cpp
gamberoillecito: You are right, that's the issue of using AI and not being careful reviewing it. I moved all the logic to library_nav

Pala_One_2_1/src/storage/library_menu_order.cpp
ncont: Ah sorry. i mixed it up. you did all right.

ncont:
i can already envision someone asking "why is entry XYZ is not visible" and "how to add an entry": Could you add a Readme section for that kind of situation?

Pala_One_2_1/src/ui/screens/library_screen.cpp
```diff
@@ -21,6 +22,7 @@
 #include "src/ui/reader_actions.h"
 #include "src/ui/lock.h"
 #include "src/ui/sleep.h"
+#include "src/pure/library_nav.h"
```
ncont: Already present in Line 5.

Pala_One_2_1/src/pure/library_nav.cpp
```diff
@@ -8,6 +8,26 @@ static void addBookEntry(int bookIdx, int depth,
...
+    case LIB_ENTRY_APPS:
+    case LIB_ENTRY_STATISTICS:
+    case LIB_ENTRY_ABOUT:
+    case LIB_ENTRY_UPDATE:
+    case LIB_ENTRY_UPLOAD:
+      return true;
+    case LIB_ENTRY_FOLDER:
+    case LIB_ENTRY_BOOK:
+      if (includeSystemEntries) {
+        return true;
```
ncont, line 24: This statement is a implicit fallthrouh to default. Add either `[[fallthrough]];` or an explicit "return false". 
See: https://en.cppreference.com/cpp/language/attributes/fallthrough for more details

gamberoillecito:
> i can already envision someone asking "why is entry XYZ is not visible" and "how to add an entry": Could you add a Readme section for that kind of situation?

I added a small paragraph explaining the change. Please let me know if you had anything different in mind

PR #69, opened 2026-05-28
Web UI: replace the String-built HTML server with a vanilla-JS SPA + JSON API (#38)

gnatpat:
## What this changes

Replaces the device's web interface. The old server built every page by concatenating HTML strings. This swaps that for a single-page app backed by a JSON API.

- **One SPA, served from `/`** — gzipped out of PROGMEM. The browser calls a JSON API under `/api/*` for state and actions; the firmware no longer renders HTML.
- **Every screen ported**: files & books, bookmarks (list + per-bookmark view + export), reading list, settings, screensavers (bitmap editor + rotation slots), factory reset, and an in-browser book reader with find + jump-to-offset.
- **Vanilla JS, no framework**

## How it works

- `webui/` holds the editable source (`index.html`, `styles.css`, `app.js`, `screens/*.js`, `i18n/*.js`).
- `scripts/build_webui.py` inlines the CSS/JS, gzips it, and emits a committed PROGMEM header at `Pala_One_2_1/src/web/generated/webui.gz.h`. It runs as a PlatformIO pre-script, so `pio run` always regenerates it; the header is committed so Arduino IDE can compile without the Python step.
- Each screen is a JS module plus a JSON endpoint pair in `src/web/api_*.cpp` (`api_files`, `api_bookmarks`, `api_list`, `api_settings`, `api_screensavers`, `api_reset`).
- Non-JSON byte transfers keep existing dedicated routes: `/upload` + `/upload-app` (multipart), `/screensavers/{thumb,download,upload}` (raw bitmap I/O), `/readbook-text` + `/jumpoffset` (book reader).
- `scripts/dev_server.py` serves `webui/` locally and either proxies `/api/*` to a real device (`--device <ip>`) or returns mock fixtures, so the UI can be developed in a browser without flashing.

## i18n

Web strings moved out of C++ and into `webui/i18n/{en,es}.js`. Device-display strings stay in `src/lang/*.h`. The SPA picks its locale from `/api/info` (the compile-time `LANG_*` the firmware was built with), with a browser-language fallback and a `localStorage` override.

## For maintainers going forward

- **Don't hand-edit `webui.gz.h`**! It's generated. Edit `webui/` and let `pio run` regenerate, or run `python scripts/build_webui.py` (Arduino IDE).
- **A CI check** (`.github/workflows/webui-check.yml`) runs on every PR and fails if the committed bundle is stale. `build_webui.py` produces deterministic output across machines, so the check won't false-positive.
[...]

PR #139, opened 2026-07-27
Custom gesture intervals

gamberoillecito:
This PR gives the users the ability to tune how clicks are characterized by changing the timing thresholds from the web UI.

I tried thinking of ways the user could inadvertently make the device unusable by setting improper times and I tried to prevent them by either defining sensible upper and lower bounds or by software checks, please let me know if you find other edge cases.


Pala_One_2_1/src/storage/click_timings.cpp
```diff
@@ -0,0 +1,157 @@
+#include "src/storage/click_timings.h"
+
+#ifdef ARDUINO
```
ncont, line 3: you can remove the ifdef since we state will handle it as an abstraction layer.

Pala_One_2_1/src/lang/en.h
```diff
@@ -369,6 +369,23 @@
...
 #define D_WEB_BUTTONS_LEGACY_HINT "If enabled, mapping single, double and triple clicks has no effect."
+
+// Click timing section.
+#define D_WEB_TIMINGS_HEADING        "Click timing"
+#define D_WEB_TIMINGS_INTRO          "Tune the behavior for clicks and hold gestures."
+#define D_WEB_TIMINGS_GAP_LABEL      "Click gap"
+#define D_WEB_TIMINGS_GAP_HINT       "How long a pause between releases may last before the current click burst is committed."
+#define D_WEB_TIMINGS_SEQUENCE_LABEL "Click sequence"
+#define D_WEB_TIMINGS_SEQUENCE_HINT  "Maximum total time allowed for a multi-click burst, measured from the first release."
+#define D_WEB_TIMINGS_LONG_LABEL     D_WEB_BUTTONS_LONG " threshold"
```
ncont, line 416: I would prefer not to write the define names because the user can not see them and wont understand the names.

Pala_One_2_1/src/storage/click_timings.cpp
```diff
@@ -0,0 +1,157 @@
...
+static constexpr uint32_t kGapMin = 75;
+static constexpr uint32_t kGapMax = 5000;
+static constexpr uint32_t kSequenceMax = 10000;
+static constexpr uint32_t kLongMin = 50;
+static constexpr uint32_t kLongMax = 10000;
+static constexpr uint32_t kVeryLongMax = 20000;
+static constexpr uint32_t kDebounceMin = 0;
+static constexpr uint32_t kDebounceMax = 100;
+
+static uint32_t s_gapMs = MAX_CLICK_GAP_MS;
```
ncont: you should move the constants to this header file as they are not compile time constants anymore but UI defaults. And it it is a bit easier to have them in the very same spot as the boundary limits.

Pala_One_2_1/src/storage/click_timings.h
```diff
@@ -0,0 +1,37 @@
...
+void resetMaxClickGapMs();
+void resetMaxClickSequenceMs();
+void resetLongMs();
+void resetVeryLongMs();
+void resetDebounceMs();
+void resetToDefaults();
+
+}  // namespace ClickTimings
+
+#endif  // PALA_STORAGE_CLICK_TIMINGS_H
```
ncont: Missing new line

Pala_One_2_1/src/storage/click_timings.cpp
```diff
@@ -0,0 +1,157 @@
...
+  s_gapMs = MAX_CLICK_GAP_MS;
+  s_sequenceMs = MAX_CLICK_SEQUENCE_MS;
+  s_longMs = LONG_MS;
+  s_veryLongMs = VERY_LONG_MS;
+  s_debounceMs = DEBOUNCE_MS;
+  normalize();
+  saveSettings();
+}
+
+}  // namespace ClickTimings
```
ncont: Missing new line

test/test_click_timings.cpp
```diff
@@ -0,0 +1,45 @@
...
+}
+
+TEST_CASE("ClickTimings — long and very-long stay ordered") {
+  ClickTimings::resetToDefaults();
+  ClickTimings::setLongMs(1500);
+  ClickTimings::setVeryLongMs(1000);
+
+  CHECK_EQ(ClickTimings::longMs(), 1500u);
+  CHECK_EQ(ClickTimings::veryLongMs(), 1501u);
+}
```
ncont, line 45: Missing new line

ncont:
While i wrote the comments above, claude reviewed as well and found the following: 

```
Issues

1. Negative/garbage form input clamps to the maximum, not the minimum
src/web/settings.cpp:520,526,532,538,544 do (uint32_t)server.arg("tim_gap").toInt() before handing the value to ClickTimings::setXxx(). toInt() returns a signed value; casting a negative number (e.g. a user typing -5, or a crafted POST bypassing the HTML min) to uint32_t wraps to a huge unsigned value (~4.29B), which clampU32 (src/storage/click_timings.cpp:31-35, used by the setters at lines 92/98/103/109/114) then clamps to the field's max bound instead of its min. So an invalid low value silently becomes the largest allowed value — e.g. a bad debounce submission jumps to 100ms, a bad long-press submission jumps to 10s. Not device-bricking, but backwards from the obviously-intended behavior, and undermines the "prevent the user from making the device unusable" goal stated in the PR description. Worth guarding against negative input before the cast (or clamping on the signed value first).

2. Reset button loses its "secondary" styling — visual hierarchy breaks
src/web/settings.cpp:438: class='secondary' — missing the btn class. Every other button/link in the codebase that wants the muted/secondary look uses class='btn secondary' (see chrome.cpp:152, bookmarks.cpp:77, files.cpp:133, etc.), because the CSS rule is a compound selector .btn.secondary (chrome.cpp:71) layered on top of the base button,.btn{...} rule (chrome.cpp:70). Since plain <button> already gets the base rule, this won't look unstyled — but it will render with the same primary color as the "Save timings" button, so all five per-field Reset buttons will visually compete with the actual Save action instead of reading as secondary. Suggest adding btn to the class list.

Code smells

3. defaultFieldValue parameter is dead
src/web/settings.cpp:77-86 (forward decl), 411-420 (definition), 446 ((void)defaultFieldValue;). Every one of the five call sites (357-366) passes a value for it, but the function immediately discards it. Looks like leftover from a refactor — either it should be used (see #4) or dropped from the signature.

4. min/max shown in the number input don't match the real server-side bounds
appendTimingField (src/web/settings.cpp:411-447) hardcodes max='60000' for every field (line 433) regardless of which constant actually governs it — e.g. debounce is really capped at 100 (kDebounceMax, click_timings.cpp:23), gap at 5000, long/sequence at 10000, very-long at 20000. The min passed in per call site is similarly a static default (e.g. 0 for gap) rather than the real kGapMin = 75, and for sequence/very-long the true minimum is dynamic (tied to the current gap/long value) so no static number can represent it anyway. Net effect: the client-side hint is decorative and can mislead a user into thinking a value was accepted at face value when it gets silently clamped after submit. This unused defaultFieldValue param (#3) suggests the author may have meant to pass a real per-field max through it and didn't finish wiring it up.
```

Hope it helps. Thanks again for the contribution !

Pala_One_2_1/src/lang/en.h
gamberoillecito: Hey, thank for all the feedback, could you clarify what the user cannot see/understand?

Pala_One_2_1/src/storage/click_timings.cpp
gamberoillecito: You mean moving e.g. kGapMin from click_timings.cpp to click_timings.h?

Pala_One_2_1/src/storage/click_timings.cpp
```diff
@@ -0,0 +1,245 @@
+#include "src/storage/click_timings.h"
+
+#include "src/state.h"
```
Higgs2209, line 4: I think this is what is breaking the build. `state.h` includes `<Arduino.h>` unconditionally.
Both includes are only needed inside the #ifdef ARDUINO bodies further down, so:

```
#ifdef ARDUINO
#include "src/state.h"
#include "src/storage/preferences_store.h"
#endif
```

Pala_One_2_1/src/storage/click_timings.cpp
```diff
@@ -0,0 +1,245 @@
...
+}
+
+static void saveToStore(KeyValueStore& kv) {
+  normalize();
+  for (const TimingSetting* setting : kAllSettings) {
+    saveSetting(kv, setting);
+  }
+}
+
+static void setSetting(const TimingSetting& setting, uint32_t value) {
```
Higgs2209, line 119: `setSetting` clamps against minValueFor, which is dynamic for the dependents — so lowering` veryLong` below long works, and the test covers that. `setLongMs(2500) `with veryLong at 2000 leaves them inverted; longSetting is MinKind::Fixed and nothing revisits veryLong. resetVeryLongMs() is worse, resetSetting doesn't clamp at all, so it drops veryLong straight to 2000 under a long of 5000.

On device both get papered over because the setters call saveSettings(), which normalizes on the way out. On host that overload is compiled out (the saveSettings(kv) one isn't, which is why the round-trip test looks fine), so the setters and the firmware behave differently. That's the bit I'd want fixed regardless.

Inverted, [long, veryLong) is empty and HoldLong can't fire — classifyHoldRelease(3000, 0) gives VeryLong. With no device-side recovery that's a lockout if a required action sits on Long.

normalize() at the end of both should do it.

Higgs2209 (changes requested):

Pala_One_2_1/src/storage/click_timings.cpp
gamberoillecito: Thanks for pointing that out! I've been struggling to think of all these cases. I'll fix this as soon as possibile

gamberoillecito: Hi, just to make sure, you meant adding `normalize()` before `saveSettings()` in `setLongMs` and `setVeryLongMs`?

Pala_One_2_1/src/lang/en.h
ncont: I can not even remember. :D sorry for the late reply. Life is busy rn

Pala_One_2_1/src/storage/click_timings.cpp
ncont: No I would move the configs from config.h think of someone would need to change the basic value. Than he would need to go to config.h and to this file to change the boundaries. I would place it all in one file.

Pala_One_2_1/src/storage/click_timings.cpp
Higgs2209: One level down; in the private `setSetting()` and `resetSetting()` helpers rather than the named public setters.

The gap/sequence pair has exactly the same coupling (`sequenceSetting` is `MinKind::Gap`, `gapSetting` is `MinKind::Fixed`), so `setMaxClickGapMs(3000)` with sequence at 550 inverts that pair too. And the reset entry points don't go through the setters at all; `resetVeryLongMs()` under a long of 5000 drops very-long to 2000, and `resetMaxClickSequenceMs()` with gap at 3000 drops sequence to 550. Both leave the pair inverted.

So if you patch just `setLongMs` and `setVeryLongMs` you'd cover one of the four cases that actually need it, and `setVeryLongMs` is the one that doesn't, it already clamps correctly through `minValueFor`.

In the helpers it's two lines and covers all ten public entry points:

```cpp
static void setSetting(const TimingSetting& setting, uint32_t value) {
  *setting.value = clampU32(value, minValueFor(setting), setting.maxValue);
  normalize();
}

static void resetSetting(const TimingSetting& setting) {
  *setting.value = setting.defaultValue;
  normalize();
}
```

`normalize()` is five comparisons and unchanged , and `kAllSettings` already lists drivers before dependents, so one pass converges. Since the setters call `setSetting` first and `saveSettings()` after, it lands before the write anyway — same order you were describing.

One thing worth deciding deliberately while you're in here: clamping can only raise the dependent, never lower the driver. So `setVeryLongMs(100)` with long at 5000 silently gives you 5001 rather than what was typed. The `min` attribute on the input normally prevents that, but the reset buttons carry `formnovalidate`, so it's reachable. Might be worth a hint in the UI or a comment saying it's intentional.

gamberoillecito: Hello, thanks again for your time. I implemented your suggestions, let me know if there is something that I missed. Regarding your last point about feedback to the user how would you implement that? Something like what I did with the customizable click patterns (e.g. a javascript `alert` that informs the user of the change)? Some form of client side validation would be extremely useful here but writing js inside cpp strings is a nightmare. If #69 gets merged we may be able to do something cleaner. I'm open to suggestions nonetheless.

On the same topic, do you think that setting `veryLong` to `long + 1` makes sense? It is formally correct but the user would need to have a 1ms accuracy to be able to actually perform `long`, which could result in being locked out of the device anyway. Maybe we should set it to `veryLong + 100` or something else?

gamberoillecito: P.S. github is telling me that you requested a change below but I cannot see what this change is (I don't have a lot of experience with opening PRs). Should I Re-request a review or do something else?

Pala_One_2_1/src/storage/click_timings.cpp
```diff
@@ -0,0 +1,273 @@
...
+
+static constexpr uint32_t kGapMin = 75;
+static constexpr uint32_t kGapMax = 5000;
+static constexpr uint32_t kSequenceMax = 10000;
+static constexpr uint32_t kLongMin = 50;
+static constexpr uint32_t kLongMax = 10000;
+static constexpr uint32_t kVeryLongMin = kLongMin + 1;
+static constexpr uint32_t kVeryLongMax = 20000;
+static constexpr uint32_t kDebounceMin = 0;
+static constexpr uint32_t kDebounceMax = 100;
```
Higgs2209: This is the block that is breaking `/ctest`.
These are static inside the `namespace ClickTimings`, meaning they are internal to this `.cpp` and the tests cannot see them. 

Moving them to `click_timings.h` fixes this and is what ncont was after I beloieve.

Higgs2209:
```suggestion
```

Pala_One_2_1/src/storage/click_timings.h
```diff
@@ -0,0 +1,50 @@
+#ifndef PALA_STORAGE_CLICK_TIMINGS_H
+#define PALA_STORAGE_CLICK_TIMINGS_H
+
+#include <stdint.h>
+
+#include "src/config.h"
+#include "src/storage/kv_store.h"
+
+namespace ClickTimings {
```
Higgs2209, line 9:
```suggestion
namespace ClickTimings {

// ---------------------------------------------------------------------------
//  Defaults and bounds.
//
//  These are what the UI resets to and the range it will accept. They are no
//  longer the values the classifier reads — input.cpp and hold_gesture.cpp go
//  through the accessors below, which return the live, user-tuned values.
// ---------------------------------------------------------------------------

// Max silence after the most recent release before we commit a click sequence.
// The press-in-progress gate in input.cpp's trailing-silence check means this
// effectively bounds "dead time between release and the next press" — once a
// press starts, the commit pauses until that release lands. So the user can
// take this long to *start* their next click; the press itself can take as
// long as it wants (up to the long-press threshold).
static constexpr uint32_t kDefaultGapMs = 175;

// Max total duration of a multi-click sequence, measured from the first release.
// Conceptually: "the whole multi-click input has to complete within this window."
// Combined with kDefaultGapMs this caps both per-gap and overall sloppiness.
static constexpr uint32_t kDefaultSequenceMs = 550;

static constexpr uint32_t kDefaultLongMs = 850;

// Hold this long (without a preceding click) and the classifier emits
// VeryLong instead of Long. Long and VeryLong — plus the click-then-hold
// chord — are independently bindable to reader actions (bookmark / lock /
// menu / none) via the web settings UI.
static constexpr uint32_t kDefaultVeryLongMs = 2000;

static constexpr uint32_t kDefaultDebounceMs = 14;

static constexpr uint32_t kGapMin = 75;
static constexpr uint32_t kGapMax = 5000;
static constexpr uint32_t kSequenceMax = 10000;
static constexpr uint32_t kLongMin = 50;
static constexpr uint32_t kLongMax = 10000;
static constexpr uint32_t kVeryLongMin = kLongMin + 1;
static constexpr uint32_t kVeryLongMax = 20000;
static constexpr uint32_t kDebounceMin = 0;
static constexpr uint32_t kDebounceMax = 100;
```

Higgs2209, line 9: Moved this block over from the `.cpp` (kept your original comments on the three values).

All values and bounds are unchanged (175 / 550 / 850 / 2000 / 14). I just renamed the defaults to `kDefault*` since the classifier reads them through accessors now, so names like `LONG_MS` were getting misleading for reset values.

A few call sites need to be updated in the same commit to keep CI green:

* **`.cpp` updates:** Swap in the new `kDefault*` names in the `s_*` initializers (~lines 50–54), `TimingSetting` tables (~72–92), and `TimingSettingSpec` table (~220).
* **Test files:** Qualify them with the namespace (e.g. `ClickTimings::kDefaultLongMs` instead of `LONG_MS`) in `test_click_timings.cpp:10-14` and throughout `test_hold_gesture.cpp`. Leaving them bare in global scope would invite collisions down the line.
* **Unused include:** `#include "src/config.h"` on line 6 isn't needed here anymore. We can drop it, but double-check the firmware build after since it transitively pulled `lang.h`.

Verified locally with these changes: 127 tests, 385 checks passing (0 failures).

Pala_One_2_1/src/web/settings.cpp
```diff
@@ -455,6 +528,20 @@ static bool applySettingsForm() {
...
+static void applyTimingSettingsForm() {
+  if (!server.hasArg("timings_form")) return;
+  const ClickTimings::TimingSettingSpec* timingSettings = ClickTimings::timingSettings();
+  for (uint8_t i = 0; i < ClickTimings::timingSettingsCount(); i++) {
+    const TimingUiField& uiField = kTimingFields[i];
+    const ClickTimings::TimingSettingSpec& timing = timingSettings[i];
+    if (server.hasArg(uiField.resetName)) {
+      timing.reset();
+    } else if (server.hasArg(uiField.nameId)) {
+      timing.set((uint32_t)server.arg(uiField.nameId).toInt());
```
Higgs2209:
```suggestion
} else if (server.hasArg(uiField.nameId)) {
      // toInt() is signed; casting a negative straight to uint32_t wraps to
      // ~4.29e9, which clampU32 then pins to the field's max instead of its
      // min. Floor it first so bad input lands on the minimum.
      long raw = server.arg(uiField.nameId).toInt();
      timing.set(raw < 0 ? 0u : (uint32_t)raw);
    }
```

Higgs2209: `toInt()` returns a signed value, so casting a negative number to `uint32_t` wraps it to ~4.29 billion. That makes `clampU32` pin the value to `max` instead of `min`—e.g., `-5` in Debounce ends up storing 100ms, and `-5` in Long silently stores 10,000ms.

HTML `min` validation would usually catch this, but the Reset buttons use `formnovalidate` and submit the entire form, so invalid inputs in other fields slip right through.

This goes back to @ncont's point from earlier

Higgs2209 (changes requested):
Thanks for sticking with this—the normalization fix in `dfd16c1` properly resolves the pair inversion edge cases across all four entry points. 

I've left two inline suggestions on the diff:
1. **`click_timings.h:9`** — Moving the bounds and default values into the header with `kDefault*` names so tests can link against them.
2. **`settings.cpp:540`** — Parsing the form input to a signed `long` before casting, preventing negative values from wrapping to `~4.29B` and pinning to max on submit.

The remaining items to address before merge:

**1. Architecture & Test State Isolation**
* **Decoupling `pure/`:** Right now, `src/pure/hold_gesture.h` includes `src/storage/click_timings.h` (pulling in `kv_store.h`), breaking the contract that `pure/` modules stay free of storage/hardware dependencies. The cleanest fix is passing the threshold values as arguments to `classifyHoldRelease()` and `classifyHoldInProgress()` directly at the two call sites in `src/hal/input.cpp`
* **Test Isolation:** Passing thresholds directly completely fixes the state leakage between test files. Currently, `test_hold_gesture.cpp` passes only because of link order; if run after `test_click_timings.cpp` (which mutates globals to 1500/1501), 5 test cases fail. If you prefer not to change the function signatures right now, add `ClickTimings::resetToDefaults()` at the start of each `TEST_CASE` in `test_hold_gesture.cpp` instead.

**2. Safety Bounds & Minimum Separation**
To ensure users cannot configure values that soft-lock their hardware or break gestures

* **`kLongMin = 300`** (at `50ms`, regular `60–150ms` taps register as holds, locking users out of the button-activated UI)
* **`kDebounceMax = 50`** (keeps debounce under `kGapMin = 75` so multi-click sequences don't become structurally impossible)
* **Replace `+1` with explicit minimum separation constants:**
  ```cpp
  // Minimum separation to prevent windows collapsing below human reaction thresholds
  static constexpr uint32_t kMinHoldSeparationMs = 250;
  static constexpr uint32_t kMinSequenceSeparationMs = 100;

gamberoillecito:
Hey, hopefully I've addressed all your comments!

Pala_One_2_1/src/web/settings.cpp
```diff
@@ -455,6 +528,26 @@ static bool applySettingsForm() {
...
+    } else if (server.hasArg(uiField.nameId)) {
+} else if (server.hasArg(uiField.nameId)) {
+      // toInt() is signed; casting a negative straight to uint32_t wraps to
+      // ~4.29e9, which clampU32 then pins to the field's max instead of its
+      // min. Floor it first so bad input lands on the minimum.
+      long raw = server.arg(uiField.nameId).toInt();
+      timing.set(raw < 0 ? 0u : (uint32_t)raw);
+    }
+    }
+  }
```
Higgs2209, line 546: Suggestion error on my part here - sorry still learning github!

`else if` is duplicated. Reset still works, but Save discards every typed value

```suggestion
    if (server.hasArg(uiField.resetName)) {
      timing.reset();
    } else if (server.hasArg(uiField.nameId)) {
      // toInt() is signed; casting a negative straight to uint32_t wraps to
      // ~4.29e9, which clampU32 then pins to the field's max instead of its
      // min. Floor it first so bad input lands on the minimum.
      long raw = server.arg(uiField.nameId).toInt();
      timing.set(raw < 0 ? 0u : (uint32_t)raw);
    }
  }
}
```


Worth checking on device once the change has been made. Put a value into each of the feilds, save and then reload the page and make sure that they all come back as what was typed.

Everything should be ready to merge once this has been fixed.

Sorry for so much back and forth!

Higgs2209 (commented):
This one is my mistake, sorry for the hassle.

Pala_One_2_1/src/web/settings.cpp
gamberoillecito: Hello, I must admit that I just checked that it was compiling correctly... I will apply your suggestions and upload to the device to make sure that it still works

gamberoillecito: I tested on device and it seems to work fine!

Higgs2209 (approved):

PR #144, opened 2026-08-10
Localize reader menu, HTTP error bodies, and embedded JS strings for language swaps

tomascorreag:
Extracts the remaining hardcoded English UI strings into the `D_*` language tables and adds the `es_la` translations. Three commits, one layer each.

### Reader menu overlay + toasts
`reader_menu.cpp` had literal `"Reading"`, `"Page %d of %d  (%d%%)"`,
`"Statusbar: %s"`, the three statusbar mode labels, and the footer hint.
`screen_settings.cpp` had `"Flipped"`; `statistics.cpp` built the streak toast by String concatenation. The streak toast is now an `snprintf` into a `char[48]` so the day number stays a format argument.

### Plain-text HTTP error bodies
`find.cpp`, `screensavers.cpp`, `files.cpp`, and `apps_upload.cpp` sent literal 4xx/5xx bodies. These surface as page content when a user hits a bad URL, so they're user-facing. Existing convention preserved: lowercase strings are argument-validation failures, capitalized ones are operation failures. The screensaver size error now formats
`Screensavers::SCREENSAVER_BYTES` instead of hardcoding `3904`.

### Embedded JS status strings + theme toggle
The find/jump and screensaver-editor `PROGMEM` script blocks built status text by string concatenation, which pins argument order to English word order. Added `window.palaFmt` to the theme script in `chrome.cpp` — a positional `%1..%9` formatter — so a translation can reorder arguments. It lives there because that script is emitted on every page.

Strings substituted into single-quoted JS literals inherit the existing no-apostrophe / no-backslash rule from `lang.h`; both language tables were checked against it. The header theme toggle (`title=`, `Dark mode`, `Light mode`) was also still hardcoded.

### Verification

- `pio run -e wireless-paper-v1_2-en` — success
- `pio run -e wireless-paper-v1_2-es` — success (flash 82.8%)
- `pio check -e wireless-paper-v1_2-en --fail-on-defect low` — no defects
- `en.h` / `es_la.h` `#define` sets are identical apart from include guards

Note that `build.yml`'s matrix only covers the `-en` environments, so a Spanish-only build break would not be caught in CI. The `-es` build above was run locally to cover that.

Host tests were not run locally (no working host toolchain on my machine); none of the sources in `test/CMakeLists.txt` are touched by this branch.


Higgs2209:
Well-scoped change, and the inline comments on argument order and quoting constraints are a genuine improvement.

Approving.

Issue #104, opened 2026-06-11
Option to remove lock icon when device is locked


Higgs2209:
A few users have requested the ability to diable the lock icon via a webUI toggle.

Issue #100, opened 2026-06-07
Add Icons to Menu


ncont:
Some Task should display a icon in the menu header: 
| Scenarion | Icon | Reason |
|  --- | --- | --- |
| Wifi is on | Wifi Icon |  User should be able to see that it might drain battery |
| Sleep is prevented | crossed moon | User should be able to see that it might drain battery  |

More ideas are welcome!

gamberoillecito:
Maybe not totally related but an indicator that shows if the user can go to the next/previous page of a multi page menu (not for books).

tomascorreag:
I can take this one

PR #145, opened 2026-08-10
Header status icons, menu scroll indicator, and label truncation

tomascorreag:
Addresses #100. Also picks up the multi-page menu indicator @gamberoillecito suggested in that thread, plus ellipsis on truncated menu labels.

## Status icon tray

New `src/ui/icons.{h,cpp}`: a right-anchored tray that packs icons leftward from the battery. Two icons, each toggleable from the Web UI (`cfg_ico_wifi`, `cfg_ico_sleep`, both default on).

- **Wi-Fi on** — new `wifiActive()` accessor in `hal/wifi.{h,cpp}`, since nothing outside the upload/update screens could previously tell whether the radio was up.
- **Sleep blocked** — new `Sleep::inhibited()`: true when the current screen declines sleep (UploadScreen always, UpdateScreen while its session is up) or a USB host holds the Improv CDC port open.

There was no icon abstraction before this — icon code was spread across `hal/battery.cpp`, a file-static padlock in `ui/sleep.cpp`, and inline drawing in `ui/reader.cpp`. The tray and its per-glyph convention are built so #104's hideable lock icon can reuse them rather than adding a fourth pile.

`Sleep::inhibited()` also now backs the two main-loop sleep gates, so the icon can't drift out of sync with the behaviour it reports. That part is an exact rewrite. The locked-idle re-sleep gate deliberately still checks only `allowSleep()` — folding the provisioning term in would stop a locked device returning to its screensaver whenever USB is attached.

## Scroll indicator

`drawScrollableList` draws up/down arrows when a list has items outside the visible window, with a reserved 8px right gutter so long labels can't collide with them. Reserved only while a list actually scrolls. All five menu screens inherit it. Because navigation wraps, the arrows mean "there are items outside this window", not "a next page exists".

## Label truncation

`truncateWithEllipsis` in `pure/text_util.cpp` (host-tested, 6 new cases), applied inside `drawMenuRow` so every list screen gets it from one place.

## Worth a maintainer decision: the Wi-Fi icon

Wi-Fi is strictly session-scoped. `wifiStaBegin()` / `wifiBeginAccessPoint()` are only called from the upload and update screens, `wifiEnd()` does `WiFi.mode(WIFI_OFF)`, and nothing brings the radio up at boot — saved credentials exist so upload mode can use STA instead of SoftAP (`storage/wifi_creds.h`, and `lang/en.h` says as much to the user).

So the Wi-Fi icon only ever renders on the upload/update headers, above a body that already prints the connection URL. I built it because it's what the issue asks for and the missing accessor was a real gap, but I'd understand dropping it. The sleep icon is the one that earns its place: a USB host silently blocks both deep and light sleep while you browse menus normally.

## Behaviour changes reviewers will notice

- Header titles now clamp with an ellipsis instead of running underneath the battery. This applies on every screen, including with both icons off. Required — without it the tray's self-clear chops a long title mid-glyph at the larger body sizes.
- Library, Apps, Bookmark list and Book select previously let long labels clip at the screen edge. They now ellipsize.
- The scroll gutter slightly narrows rows on scrolling lists, so a To-Do label that fit on one line may now wrap to two.
- AboutScreen opts out of the tray (`drawSectionHeader(..., false, false)`) — its extended battery block occupies the same band.

## Verification

- `pio run` on `wireless-paper-v1_2-en`, `wireless-paper-v1_1-en` and `wireless-paper-v1_2-es`. The `-es` build is the only check on `es_la.h` key parity, since CI builds `-en` only.
- `pio check -e wireless-paper-v1_2-en --fail-on-defect low` — no defects.
- `ctest` host suite: 124 tests, 0 failures.


Pala_One_2_1/src/ui/widgets.cpp
```diff
@@ -55,19 +74,27 @@ void drawCenter(const char* a, const char* b) {
...
   if (drawBattery){
     drawBatteryTopRight();
   }
 #endif
+  if (drawIcons) {
+    right = Icons::drawStatusTray(Icons::trayRightEdge(drawBattery));
+  }
+
+  u8g2.setCursor(MARGIN_X, yTitle);
+  u8g2.print(fitLabelWithEllipsis(String(title), right - MARGIN_X - 4).c_str());
```
Higgs2209, line 100: This prints the title in the wrong font. `drawBatteryTopRight()` -> `drawBattery()` calls `Font::useUiSmall()` at battery.cpp:288 and never restores it, so by the time we reach here the current face is `u8g2_font_6x10_tf`. Both the `getUTF8Width` calls inside `fitLabelWithEllipsis` and the print itself run under it, while yTitle on line 80 came from bold's ascent.

Drawing the title first used to hide this. Now it lands on 11 of the 12 `drawSectionHeader` callers, since drawBattery defaults true and only Device opts out.

The part users will notice: `s_bold` follows the size and family from `pickFaces`, `s_uiSmall` is fixed 6x10. A device set to size 14 gets 6x10 headers, so this takes back some of what #134 added.

A `Font::useBold();` right before the `setCursor` fixes it. The `Font::useUiSmall()` in drawBattery is vestigial anyway now that the percentage text is gone, worth deleting on its own, but I'd still want the re-select here rather than having the header depend on what the indicators leave set.

Pala_One_2_1/src/hal/battery.cpp
```diff
@@ -272,6 +272,16 @@ void drawBolt(int battX, int battY, int battH, int spacing)
 const int iconW = 18;
 const int iconH = 9;
 
+// Widest adornment drawn to the LEFT of the battery body. The charging bolt
+// (drawBolt) reaches battX - spacing(2) - iconH/2(4) - 1 = battX - 7; the
+// low-battery exclamation (drawExclamation) only reaches battX - 4. Anything
+// laid out beside the battery has to clear the larger of the two.
+static const int kTopRightAdornW = 7;
```
Higgs2209: `drawBolt` draws each of its three segments twice, at `x - i` for i in {0,1}, so the leftmost pixel is `d_x - 1` = `battX - 8`. The arithmetic in the comment stops at `d_x`.

Nothing overlaps today -> `kTrayEdgeGap` of 3 puts the tray's rightmost column at `battX - 11`, leaving 2 clear pixels. This is really about the comment, which reads as the authoritative geometry for whoever adds the next glyph.

test/test_text_util.cpp
```diff
@@ -134,3 +134,71 @@ TEST_CASE("compactText streaming: no word merging across boundary") {
...
+    n++;
+    i += charLen;
+  }
+  return n * 6;
+}
+
+TEST_CASE("truncateWithEllipsis leaves a string that already fits untouched") {
+  CHECK_EQ(truncateWithEllipsis("abc", 18, measure6PerCodepoint), String("abc"));
+  // Exactly at the limit is still a fit.
+  CHECK_EQ(truncateWithEllipsis("abc", 18, measure6PerCodepoint), String("abc"));
```
Higgs2209: This isi the same call as line 160. Did you mean a larger width here?

Higgs2209 (changes requested):
Nice work.

Can you please rebase before this goes in? You're 6 commits behind dev and #144 has landed since you forked. It merges clean, but I'd rather be reviewing the result that actually ships.

On the Wi-Fi icon, your read matches mine, and I'd drop it. `wifiActive()` is worth keeping regardless, and what #104 needs is the tray plus the drawing convention, not a second glyph already in place. Not blocking if you'd rather keep it. The sleep icon clearly earns its slot.

tomascorreag:
@Higgs2209 thanks for the review, I agree on all counts. It's all fixed in the latest commit, let me know if you have more feedback.

Issue #147, opened 2026-08-23
Files become unusable after relative path is greater than 89 chars (on device and in WebUI)


milofultz:
## Problem

You can upload files of any length, but files with a relative path longer than 89 characters can no longer be opened, deleted, or moved

## Environment

I believe I am on build `v2.1-218-g541c508`. Used the flash installer. All reproduction steps are from factory reset.

### Post upload

1. Make a file that is a total of 89 characters long: `name=""; for i in {1..85}; do name="${name}a"; done; echo "test" > "$name.txt"`
2. Upload this file to the Pala.
3. Try to delete or move the file. We expect success, but instead get a silent error (deletion) or an error response of "move failed" (move).
4. On the device, try to open the file. We expect it to open, but it instead says "Open failed, try upload again".

### After moving to folder

1. Make a file that is a total of 88 characters long: `name=""; for i in {1..84}; do name="${name}a"; done; echo "test" > "$name.txt"`
2. Upload this file to the Pala.
3. Delete the file. Note that it succeeds.
4. Upload the file again.
5. Create a folder of one character in length (e.g. "a").
6. Move the file to your new one character folder.
7. Try to delete or move the file. We expect success, but instead get a silent error (deletion) or an error response of "move failed" (move).
8. On the device, try to open the file. We expect it to open, but it instead says "Open failed, try upload again".

## Hypothesis

The absolute path on the Pala exceeds the firmware's expected filename length (128?) and then fails with no recourse but to flash the device. If we could put a check on the web UI for relative path name length, then we could block this issue and inform users on how to fix (e.g. something like `fp="$folder/$filename"; if [ ${#fp} -ge 85 ]; then echo "error: relative filepath exceeds 85 characters" ; exit`).

milofultz:
Could also be we can make the filepaths longer using `CONFIG_FATFS_LONG_FILENAMES`, but I don't know enough about this hardware to know for certain.

https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/kconfig-reference.html#config-fatfs-long-filenames

milofultz:
Because this uses LittleFs under the hood, my guess is the max is 255: https://github.com/littlefs-project/littlefs/blob/6cb4e86540eca0d9ba62500a298385c9d863c8be/lfs.h#L51

milofultz:
[...]

PR #148, opened 2026-08-23
Issue #147: Clamp filename length on upload

milofultz:
#147 : This handles errors on file upload by forcing filenames to be below the 89 character threshold.

### DIFF
--- a/Pala_One_2_1/src/config.h
+++ b/Pala_One_2_1/src/config.h
@@ -57,6 +57,7 @@ static const int SCREEN_H = 122;
 
 static const uint8_t MAX_BOOKMARKS = 12;
 static const int MAX_BOOKS = 80;
+static const int MAX_FILE_PATH = 89;  // chars, relative path from app base
 static const int MAX_FOLDERS = 32;
 static const int MAX_FOLDER_PATH = 63;  // chars, excluding null
 static const int MAX_PAGES = 10000;
--- a/Pala_One_2_1/src/pure/paths.cpp
+++ b/Pala_One_2_1/src/pure/paths.cpp
@@ -81,8 +81,7 @@ String sanitizeFolderInput(const String& raw) {
 }
 
 String sanitizeUploadedFilename(String fname) {
-  int slash = fname.lastIndexOf('/');
-  if (slash >= 0) fname = fname.substring(slash + 1);
+  fname = lastPathComponent(fname);
 
   String clean;
   clean.reserve(fname.length());
--- a/Pala_One_2_1/src/pure/paths.cpp
+++ b/Pala_One_2_1/src/pure/paths.cpp
@@ -1,3 +1,4 @@
+#include "../config.h"
 #include "paths.h"
 
 String stripTxtExt(const String& s) {
@@ -102,6 +103,10 @@ String sanitizeUploadedFilename(String fname) {
   while (clean.startsWith(".")) clean.remove(0, 1);
   if (!clean.endsWith(".txt")) clean += ".txt";
   if (clean.length() == 0) clean = "book.txt";
+  if (clean.length() >= MAX_FILE_PATH) {
+    int ext = fname.lastIndexOf('.');
+    clean = clean.substring(0, min(ext, MAX_FILE_PATH - 4)) + ".txt";
+  }
   return clean;
 }
 
--- a/Pala_One_2_1/src/pure/paths.cpp
+++ b/Pala_One_2_1/src/pure/paths.cpp
@@ -101,8 +101,8 @@ String sanitizeUploadedFilename(String fname) {
 
   clean.replace("..", "");
   while (clean.startsWith(".")) clean.remove(0, 1);
-  if (!clean.endsWith(".txt")) clean += ".txt";
   if (clean.length() == 0) clean = "book.txt";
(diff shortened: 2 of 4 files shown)


Pala_One_2_1/src/pure/paths.cpp
```diff
@@ -100,8 +100,12 @@ String sanitizeUploadedFilename(String fname) {
 
   clean.replace("..", "");
   while (clean.startsWith(".")) clean.remove(0, 1);
-  if (!clean.endsWith(".txt")) clean += ".txt";
   if (clean.length() == 0) clean = "book.txt";
+  if (!clean.endsWith(".txt")) clean += ".txt";
+  if (clean.length() >= MAX_FILE_PATH) {
+    int ext = fname.lastIndexOf('.');
+    clean = clean.substring(0, min(ext, MAX_FILE_PATH - 4)) + ".txt";
+  }
```
Higgs2209, line 108: Root cause first, since #147 was circling FS limits: it isn't the FS. This is LittleFS (state.h:15), so CONFIG_FATFS_LONG_FILENAMES doesn't apply, and your repro already shows LittleFS stores and lists the long name fine. What breaks is loadBooks() copying the path into BookInfo::path[96] via strncpy(..., 95) (library.cpp:68). /books/ is 7 chars, so the catalog holds at most an 88-char relative path — at 89 it silently truncates, and open/delete/move all use that mangled string.

This clamp still emits ≥ 89 two ways:

1. MAX_FILE_PATH - 4 keeps 85 base chars + ".txt" = exactly 89. A 90-char upload comes out at 89, and an 89-char input is rebuilt unchanged (cut at 85, re-append .txt) — which is why the tests pass.
2. With no dot in fname, ext is −1; substring(0, -1) wraps unsigned and returns the whole string (Arduino and host shim both), so the limit is skipped and .txt doubles up: 85 chars no-ext → 93 out.


And ext indexes fname but cuts clean, which is a different length after .. removal and the .txt append — ab. + 96 chars becomes 6-char ab.txt, inviting collisions (upload overwrites an existing file of the same name).

All three vanish if the clamp uses the invariant that clean already ends in ".txt" here:


```suggestion
   if (clean.length() >= MAX_FILE_PATH) {
    // clean always ends in ".txt" here; keep MAX_FILE_PATH - 5 base chars
    // so the result is at most 88 and "/books/" + name fits path[96].
    clean = clean.substring(0, MAX_FILE_PATH - 5) + ".txt";
  }
```

test/test_paths.cpp
```diff
@@ -76,6 +76,10 @@ TEST_CASE("sanitizeUploadedFilename") {
   CHECK(sanitizeUploadedFilename("weird!@#chars.txt") == "weird___chars.txt");
   // Numeric, punctuation in middle stays as filename
   CHECK(sanitizeUploadedFilename("a1-b2.txt") == "a1-b2.txt");
+  // Clamps to MAX_FILE_NAME
+  CHECK(sanitizeUploadedFilename("89_chars___________________________________________________________________________89.txt") == "89_chars___________________________________________________________________________89.txt");
+  CHECK(sanitizeUploadedFilename("90_chars____________________________________________________________________________90.txt") == "90_chars____________________________________________________________________________9.txt");
+  CHECK(sanitizeUploadedFilename("93_chars_______________________________________________________________________________93.txt") == "93_chars_____________________________________________________________________________.txt");
```
Higgs2209, line 82: These currently assert 89-char outputs. The exact length the catalog truncates and the first one locks in an 89-char pass-through. Expected values under the corrected clamp, plus the lower boundary (all literals length-verified):

```suggestion
  // Clamps to MAX_FILE_PATH - 1 (88) so "/books/" + name fits BookInfo::path[96]
  CHECK(sanitizeUploadedFilename("88_chars__________________________________________________________________________88.txt") == "88_chars__________________________________________________________________________88.txt");
  CHECK(sanitizeUploadedFilename("89_chars___________________________________________________________________________89.txt") == "89_chars___________________________________________________________________________8.txt");
  CHECK(sanitizeUploadedFilename("90_chars____________________________________________________________________________90.txt") == "90_chars____________________________________________________________________________.txt");
  CHECK(sanitizeUploadedFilename("93_chars_______________________________________________________________________________93.txt") == "93_chars____________________________________________________________________________.txt");
```

Pala_One_2_1/src/config.h
```diff
@@ -57,6 +57,7 @@ static const int SCREEN_H = 122;
 
 static const uint8_t MAX_BOOKMARKS = 12;
 static const int MAX_BOOKS = 80;
+static const int MAX_FILE_PATH = 89;  // chars, relative path from app base
```
Higgs2209, line 60: The comment says "relative path from app base" but as I understand it, the check enforces only the bare filename at upload.  A fule at max length moved into a folder will cross the limit again.

Would you rather reword this to say filename, or rename to `MAX_FILE_NAME`.

Higgs2209 (changes requested):

milofultz:
Thank you @Higgs2209, I appreciate your feedback on this. I have purely worked on software so hardware debugging/programming is very novel for me and this was super helpful. I'll get back to this later this week and work towards the real root issue. Thanks again!

Higgs2209 (approved):

